PR #5, opened 2021-09-27
Describe build hints

rshym:


lorc:
I think that hints should be moved to a separate document/page. Something like "Usage Notes".

docs/user-reference.rst
```diff
@@ -368,3 +368,39 @@ Optional parameters:
...
+* :code:`domd` 
+* :code:`fetch-domu`
+* :code:`conf-domu`
+* :code:`domu` 
+
+Inside-domain build
+^^^^^^^^^^^^^^^^^^^
+
+If you need to work inside the domain's build environment you need to specify
+corresponding build folder as a parameter to :code:`oe-init-build-env`, like :code:`build-domu` or :code:`build-domd`.
```
lorc, line 398: I think you need to mention that build folder corresponds to `work_dir` parameter in a builder configuration.

docs/usage-notes.rst
```diff
@@ -0,0 +1,37 @@
+Usage notes
+===========
+
+Component-specific targets
+--------------------------
+
+As mentioned above in parts 'Fetchers' and 'Builders', :code:`moulin`
+generates some helpful targets for components: to fetch sources,
+to create a configuration, to build a domain. For example, if we have
+product with DomU, you will have following self-explanatory targets:
```
lorc, line 10: Well, I'd rather use some more generic wording. "Product" and "DomU" may be hard to understand for some who is not familiar with XenTroops products.

I'd write something like this: "If we have components dom0, domd and domu..."

PR #9, opened 2021-10-13
doc: Add example of using different versions of same repo

rshym:


rshym:
@lorc could you please take a look?

PR #37, opened 2022-03-25
Move gzip and bmap creation from rouge to ninja rules

Deedone:
Rouge is not designed for this tasks, so move them to ninja rules.


lorc:

PR #41, opened 2022-05-23
Introducing new parameter to rouge: gpt_uuid

oleksiimoisieiev:
gpt_uuid parameter is intendend to set UUID for gpt partitions. Partition UUID can be used to mount partition from fstab regardless of the storage type.

### DIFF
--- a/docs/rouge.rst
+++ b/docs/rouge.rst
@@ -284,6 +284,7 @@ virtual machines and wish to provide VM with own GPT.
    partitions:
      boot: # partition label
        gpt_type: 21686148-6449-6E6F-744E-656564454649 # BIOS boot partition (kinda...)
+       gpt_guid: 8DA63339-0007-60C0-C436-083AC8230900 # Partition GUID
        type: empty
        size: 30 MiB
      rootfs:
@@ -298,12 +299,22 @@ includes Raw Image block.
 :code:`partitions:` section is mandatory. It defines list of
 partitions, where key is a partition label.
 
-Each partition contains definition of other block type plus optional
-(but we strongly suggest to provide it) key :code:`gpt_type:`. This
-key holds GPT Partition Type GUID. List of widely used types can be
-found on
+Each partition contains definition of other block type plus optional keys:
+
+:code:`gpt_type:` (which we strongly suggest to provide) key holds GPT Partition
+Type GUID. List of widely used types can be found on
 `Wikipedia <https://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_type_GUIDs>`_,
 for example.
+
+:code:`gpt_guid:` key sets the GPT Partition GUID. By default this GUID is generated
+automatically to ensure that every partition in the world would have unique
+identifier. But there are some cases when external software depends on exact value
+of a partition GUID. In such cases it is possible to hard-code this value. We
+strongly recommend not to use this key except for the cases when this is neccessary
+because, accoding to the page 121 of
+`Specification <https://uefi.org/sites/default/files/resources/UEFI_Spec_2_8_final.pdf>`_
+the software that makes copied of GPT-formatted disks and partitions must generate
+new Unique Partition GUID in each GPT Partition Entry.
 
 `rouge` will place partitions one after another, aligning partition
 start to 1 MiB (as per standard recommendation) and partition size to
@@ -343,6 +354,7 @@ The following example provides multiple different images:
              "initrd": "yocto/build/tmp/deploy/images/generic-armv8-xt/uInitrd"
          domd_rootfs:
            gpt_type: B921B045-1DF0-41C3-AF44-4C6F280D3FAE # Linux aarch64 root
+           gpt_guid: 8DA63339-0007-60C0-C436-083AC8230900 # Partition GUID
            type: raw_image
            image_path: "yocto/build-domd/tmp/deploy/images/machine/core-image-weston.ext4"
 
--- a/moulin/rouge/block_entry.py
+++ b/moulin/rouge/block_entry.py
@@ -37,6 +37,7 @@ class GPTPartition(NamedTuple):
     "Represents one partition in GPT"
     label: str
     gpt_type: str
+    gpt_guid: str
     start: int
     size: int
     entry: BlockEntry
@@ -51,8 +52,8 @@ class GPT(BlockEntry):
 
         for part_id, part in node["partitions"].items():
             label = part_id
-            entry_obj, gpt_type = self._process_entry(part)
-            self._partitions.append(GPTPartition(label, gpt_type, start=0, size=0, entry=entry_obj))
+            entry_obj, gpt_type, gpt_guid = self._process_entry(part)
+            self._partitions.append(GPTPartition(label, gpt_type, gpt_guid, start=0, size=0, entry=entry_obj))
 
     def size(self) -> int:
         "Returns size in bytes"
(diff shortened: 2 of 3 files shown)


moulin/rouge/block_entry.py
```diff
@@ -37,6 +37,7 @@ class GPTPartition(NamedTuple):
     "Represents one partition in GPT"
     label: str
     gpt_type: str
+    gpt_uuid: str
```
lorc, line 40: It is GUID, not UUID.

Better name would be `part_guid` to denote that this is partition GUID.

moulin/rouge/sfdisk.py
```diff
@@ -31,11 +31,15 @@ def _align(val, align) -> int:
 def _to_script(part: Any, sector_size=512) -> str:
     "Convert GPT Partition object to sfdisk script line"
 
-    return ", ".join([
+    result = ", ".join([
         f"start={_sect(part.start, sector_size)}", f"size={_sect(part.size, sector_size)}",
         f"type={part.gpt_type}", f"name={part.label}"
     ])
 
+    if part.gpt_uuid != "":
```
lorc, line 39: This should be `if part.gpt_uuid: `

moulin/rouge/sfdisk.py
```diff
@@ -31,11 +31,15 @@ def _align(val, align) -> int:
...
+        result = ", ".join([result, f"uuid={part.gpt_uuid}"])
```
lorc, line 40: Ugh.

We should rework this whole function.

Something like this will be much nicer:

```
    args = [
        f"start={_sect(part.start, sector_size)}", f"size={_sect(part.size, sector_size)}",
        f"type={part.gpt_type}", f"name={part.label}"
    ]

    if part.part_guid:
        args.append(f"uuid={part.part_guid}")

    return ", ".join(args)

```

lorc (changes requested):
Please combine both commits into one.

We want documentation to be added along with the feature in the same commit.

docs/rouge.rst
```diff
@@ -299,11 +300,13 @@ includes Raw Image block.
...
 Each partition contains definition of other block type plus optional
-(but we strongly suggest to provide it) key :code:`gpt_type:`. This
-key holds GPT Partition Type GUID. List of widely used types can be
+(but we strongly suggest to provide it) keys :code:`gpt_type:`, :code:`gpt_uuid:`.
+:code:`gpt_type:` key holds GPT Partition Type GUID. List of widely used types can be
 found on
 `Wikipedia <https://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_type_GUIDs>`_,
 for example.
+:code:`gpt_uuid:` key sets partition UUID, which can be used to access to partition
+regardless of storage type.
```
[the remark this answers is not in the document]
oleksiimoisieiev: The main idea of partition uuids and labels is that you can access to partition using some predefined name or id.
So for example: traditional way to mount partition in fstab is /dev/mmcblk0p1 or /dev/nvme...
Using PARTUUID= in fstab saves us from changing fstab each time we change storage. So we can use one image, written to /dev/mmcblk*, /dev/nvme0n*, /dev/ssd*.

Rouge already have part label, but I had to add part_uuid - because MTA has swupdate project, which does A/B update and those partition UUIDs are hardcoded to swupdate application.

moulin/rouge/block_entry.py
oleksiimoisieiev: I would disagree with that. sfdisk parameter is called uuid, fdisk entry also named as PARTUUID. The Idea of GUID is that this is global unique identifier. We can't say that partition id is globally unique.

oleksiimoisieiev: I was wrong. https://en.wikipedia.org/wiki/GUID_Partition_Table says that 0x16 position called Partition GUID on Windows manner. I will rename.

docs/rouge.rst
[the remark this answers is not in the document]
lorc: > Rouge already have part label, but I had to add part_uuid - because MTA has swupdate project, which does A/B update and those partition UUIDs are hardcoded to swupdate application.

Well, this is the good reason and good example. You can use this in the documentation:

```:code:`part_guid:` key sets partition GUID. Usually this GUID is generated automatically to ensure that every partition in the world have unique identifier. But there are some cases when external software depends on exact value of a partition GUID. In such cases it is possible to hard-code this value.```

File: moulin/rouge/sfdisk.py (hunk 2 above)
Comment (oleksiimoisieiev) [reply moulin/rouge/sfdisk.py:39]: fixed

File: moulin/rouge/block_entry.py (hunk 1 above)
Comment (oleksiimoisieiev) [reply moulin/rouge/block_entry.py:40]: fixed

File: moulin/rouge/sfdisk.py (hunk 3 above)
Comment (oleksiimoisieiev) [reply moulin/rouge/sfdisk.py:40]: fixed

File: docs/rouge.rst [hunk 5]
```diff
@@ -284,6 +284,7 @@ virtual machines and wish to provide VM with own GPT.
    partitions:
      boot: # partition label
        gpt_type: 21686148-6449-6E6F-744E-656564454649 # BIOS boot partition (kinda...)
+       gpt_guid: 8DA63339-0007-60C0-C436-083AC823090 # Partition GUID
```
lorc, line 287: Why it is one digit shorter than `gpt_type`?

docs/rouge.rst
```diff
@@ -299,12 +300,18 @@ includes Raw Image block.
 partitions, where key is a partition label.
 
 Each partition contains definition of other block type plus optional
-(but we strongly suggest to provide it) key :code:`gpt_type:`. This
-key holds GPT Partition Type GUID. List of widely used types can be
+(but we strongly suggest to provide them) keys:
```
lorc, line 303: I strongly suggest to provide `gpt_type` and strongly suggest to NOT provide `gpt_guid` unless it is strictly necessary. People just love to copy and paste. Any hard-coded partition GUID will spread across all projects. We don't want this.

docs/rouge.rst
```diff
@@ -299,12 +300,18 @@ includes Raw Image block.
...
+:code:`gpt_type:` key holds GPT Partition Type GUID. List of widely used types can be
 found on
 `Wikipedia <https://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_type_GUIDs>`_,
 for example.
 
+:code:`gpt_guid:` key sets GPT partition GUID, which can be used in fstab to mount
+partition regardless of the storage type. Details could be found on:
+`Wikipedia <https://en.wikipedia.org/wiki/GUID_Partition_Table>`_.
+
+
```
lorc, line 314: Why two line breaks?

lorc:
BTW, you can run pylint locally ;)

docs/rouge.rst
```diff
@@ -299,12 +300,18 @@ includes Raw Image block.
...
-key holds GPT Partition Type GUID. List of widely used types can be
+(but we strongly suggest to provide them) keys:
+
+:code:`gpt_type:` key holds GPT Partition Type GUID. List of widely used types can be
 found on
 `Wikipedia <https://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_type_GUIDs>`_,
 for example.
 
+:code:`gpt_guid:` key sets GPT partition GUID, which can be used in fstab to mount
+partition regardless of the storage type. Details could be found on:
```
lorc, line 311: > regardless of the storage type

What this means? I don't get it.

docs/rouge.rst
oleksiimoisieiev: Fixed

docs/rouge.rst
oleksiimoisieiev: Fixed

lorc:

lorc (approved):

PR #43, opened 2022-07-14
Add fetcher and builder for Zephyr OS

rshym:
```
Fetcher for west meta-tool
  - type: west
  - url: is optional and corresponds to `west init -m`
  - rev: is optional and corresponds to `west init --mr`
  - file: is optional and corresponds to `west init --mf`
Beside sources, fetcher also fetches required version of CMake.

Builder:
  - type: zephyr
  - board: <any supported board>
  - target: <samples/...>
  - sdk: <path to preinstalled SDK>
  - target_images: zephyr.bin

For additional info, please see
https://docs.zephyrproject.org/latest/develop/getting_started/index.html
```


rshym:
PR contains working code, that fetches and builds as expected.
But it needs some additional work, especially in part related to external tools.

moulin/builders/zephyr.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    return ZephyrBuilder(conf, name, build_dir, src_stamps, generator)
+
+
+def gen_build_rules(generator: ninja_syntax.Writer):
+    """
+    Generate Zephyr build rules for ninja
+    """
+    cmd = " && ".join([
+        "export PATH=$abs_cmake_path:$$PATH",
```
lorc, line 26: Why we can't use `cmake` from host?

moulin/builders/zephyr.py
```diff
@@ -0,0 +1,78 @@
...
+        self.src_stamps = src_stamps
+        self.build_dir = build_dir
+
+    def gen_build(self):
+        """Generate Ninja rules to build Zephyr"""
+        abs_root = os.path.abspath(self.build_dir)
+        variables = {
+            "abs_cmake_path": abs_root + "/ext_tools/cmake/bin",
+            "zephyr_base": abs_root + "/zephyr",
+            "sdk": os.path.abspath(self.conf["sdk"].as_str),
```
lorc, line 58: Do we really need to provide path to SDK? Looks like `west` can determine SDK location automatically. 
I want YAML files to be environment-agnostic. Forcing user to update YAML for own environment is not a good idea.

moulin/fetchers/west.py
```diff
@@ -0,0 +1,121 @@
...
+    ])
+    generator.rule("west_init",
+                   command=cmd,
+                   description="Create a west workspace")
+    generator.newline()
+
+    cmd = " && ".join([
+        "mkdir -p $build_dir/ext_tools",
+        "cd $build_dir/ext_tools",
+        "wget -q $cmake_url",
```
lorc, line 34: Not all distributions have wget installed by default. Instead of forcing user to install wget, it is easier to make them install cmake directly :)

moulin/fetchers/west.py
```diff
@@ -0,0 +1,121 @@
...
+                             variables={
+                                 "build_dir": self.build_dir
+                             })
+        self.generator.newline()
+
+        return update_target
+
+    def get_file_list(self) -> List[str]:
+        """Get list of files under fetcher control"""
+        return []
```
lorc, line 117: Well, it would be good to implement this. Otherwise ninja will not track changes into source code and will not rebuild zephyr, which will hinder development.

lorc (changes requested):
Please update documentation in the same patch.

Also, you can split this PR into two patches:

1. fetcher + documentation
2. builder + documentation

lorc:
Sorry for late comment, but you also need to bump up version of Moulin to ensure that `min_version` is properly used/working in YAML files.

moulin/fetchers/west.py
rshym: get_file_list() is implemented together with corresponding part of builder.
Tested successfully: I have added whitespace to one of the sources, and ninja noticed the change and ran the build.

rshym:
Documentation is added.
Version is bumped to 0.8.
All comments seem to be addressed.

lorc (approved):

rshym:
RB is added.
PR is ready to be merged.

PR #45, opened 2022-07-25
Minor fixes

rshym:
- Fix misprint in message
- Add 2022 to copyright
- `s/fetcher used/fetcher is used/` in documentation


rshym:
RB is added.
Testing is not needed.

PR #50, opened 2022-09-20
rouge: add possibility to specify size of image

rshym:
For some cases we need to have image of certain size despite sizes of internal partitions.

This PR adds `image_size` field to image descriptor. If it is not specified or zero, then image is created taking into account internal partitions.
If actual size of partitions is bigger than `image_size` - error is raised, that we can't create image of specified size.

### DIFF
--- a/moulin/rouge/block_entry.py
+++ b/moulin/rouge/block_entry.py
@@ -12,6 +12,7 @@ import itertools
 from typing import List, Tuple, NamedTuple, cast
 from tempfile import NamedTemporaryFile, TemporaryDirectory
 
+from typing import Optional
 from yaml import Mark
 from moulin.rouge import sfdisk, ext_utils
 from moulin.yaml_helpers import YAMLProcessingError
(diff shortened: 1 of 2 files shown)


rshym:
Generation of image was tested with this PR for a few cases:
- `image_size` is specified and is bigger than the actual image size - image has a specified size;
- `image_size` is specified and is smaller than the actual size - an exception is shown to the user, and no image is created;
- `image_size` is not specified - image has actual size.

rshym:
YAML with image_size looks like this:
```images:
  full:
    type: gpt
    image_size: 512 MiB
    desc: "Full SD-card/eMMC image"
    partitions:
```

rshym:
Rebased to latest `main`.

moulin/rouge/block_entry.py
```diff
@@ -49,16 +49,28 @@ class GPT(BlockEntry):
     def __init__(self, node: YamlValue):
         self._partitions: List[GPTPartition] = []
         self._size: int = 0
+        self._requested_image_size: int = 0
```
lorc, line 52: Better to make it `Optional[int]` and test for `None` rather than for `0`

lorc (changes requested):
Please update the documentation accordingly.

rshym:
Comments are addressed:
- `Optional[int]` is used and tested with and without image_size;
- documentation is updated.

rshym:
Update.
- Reference to 'Size Designation' section is added
- VS Code editor showed me a notification that 'The character U+00a0 is invisible.' So I replaced that invisible char with regular whitespace.

lorc (approved):
Please squash the two commits together.

In most cases it is good to have functional and documentation changes in the same commit because:

1) It helps to document changes made by commit

2) Both changes can be reverted at the same time

WIth that:

rshym:
RB is added.
Successfully tested with
- image_size not specified
- image_size < actual size
- image_size => actual size

Ready to be merged.

Issue #40, opened 2022-05-08
rouge: it is impossible to put files into some folder on ext4 partition


al1img:
Example;
```yaml
    partitions:
      boot_a:
        gpt_type: 21686148-6449-6E6F-744E-656564454649 # BIOS boot partition (kinda...)
        type: ext4
        size: 128 MiB
        files:
          "/my_folder/my_file": "local/file"
```

PR #55, opened 2022-09-28
Allow user to specify folders in image + minor improvements

rshym:
- Add support for destination folders;
- Allow user to specify destination folder from root
- Fix misprint in rouge.rst
- Reduce amount of output info to improve readability

### DIFF
--- a/moulin/rouge/block_entry.py
+++ b/moulin/rouge/block_entry.py
@@ -271,6 +271,10 @@ class Ext4(FileSystem):
             self._complete_init()
         with NamedTemporaryFile() as tempf, TemporaryDirectory() as tempd:
             for remote, local, _ in self._files:
+                # user can specify destination folder from root
+                # and we need to remove very first '/' for correct
+                # work of os.path.join
+                remote = remote.lstrip('/')
                 # create destination subfolders
                 remote_path_and_name = os.path.split(remote)
                 if remote_path_and_name[0]:
--- a/moulin/rouge/block_entry.py
+++ b/moulin/rouge/block_entry.py
@@ -271,6 +271,12 @@ class Ext4(FileSystem):
             self._complete_init()
         with NamedTemporaryFile() as tempf, TemporaryDirectory() as tempd:
             for remote, local, _ in self._files:
+                # create destination subfolders
+                remote_path_and_name = os.path.split(remote)
+                if remote_path_and_name[0]:
+                    # create destination subfolder
+                    os.makedirs(os.path.join(tempd, remote_path_and_name[0]), exist_ok=True)
+
                 shutil.copyfile(local, os.path.join(tempd, remote))
             tempf.truncate(self._size)
             ext_utils.mkext4fs(tempf, tempd)
@@ -285,6 +291,35 @@ class Vfat(FileSystem):
         with NamedTemporaryFile() as tempf:
             tempf.truncate(self._size)
             ext_utils.mkvfatfs(tempf)
+            # scan all remote filenames and collect the list of folders to create
+            list_for_mmd = list()
+            for remote, _, _ in self._files:
+                # remove starting '/' to avoid:
+                # - icluding different forms of same name, like "/zxc" and "zxc"
+                # - adding root folder "/" to list
+                remote_path_and_name = os.path.split(remote.lstrip('/'))
+                if remote_path_and_name[0]:
+                    # Here is explanation of what we do here.
+                    # mmd can't create chain of folders like '/a/s/d' if parent
+                    # folders (/a/s/) do not exist.
+                    # You need to create each subfolder by separate call or provide them
+                    # in incremental way, like `mmd /a /a/s /a/s/d`.
+                    # Taking into account that we work with an image file,
+                    # we need to prefix each folder with "::".
+                    # So here we split each folder into slices,
+                    # and re-assemble slices in incremental way.
+                    current_path = ""
+                    for path_slice in remote_path_and_name[0].split('/'):
+                        if path_slice:
+                            current_path += "/" + path_slice
+                            list_for_mmd.append("::" + current_path)
+            if list_for_mmd:
+                # Remove duplicates by transforming the list of strings to the set of strings.
+                # And sort strings because after 'set()' we may have 'a/s/d' before 'a/s' in some cases.
+                list_for_mmd = sorted(set(list_for_mmd))
+                # create all destination subfolders at once
+                ext_utils.mmd(tempf, list_for_mmd)
+
             for remote, local, _ in self._files:
                 ext_utils.mcopy(tempf, local, remote)
             ext_utils.dd(tempf, fp, offset)
--- a/moulin/rouge/ext_utils.py
+++ b/moulin/rouge/ext_utils.py
@@ -79,3 +79,11 @@ def mcopy(img: BinaryIO, file: str, name: str):
     args = ["mcopy", "-i", img.name, file, "::" + name]
 
     _run_cmd(args)
+
+
+def mmd(img: BinaryIO, folders: list):
+    "Create directories inside a vfat image"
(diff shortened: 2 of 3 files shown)


rshym:
Fixes issue https://github.com/xen-troops/moulin/issues/40

moulin/rouge/block_entry.py
```diff
@@ -286,6 +293,19 @@ def write(self, fp, offset):
             tempf.truncate(self._size)
             ext_utils.mkvfatfs(tempf)
             for remote, local, _ in self._files:
+
```
lorc, line 301: Remove this empty line

moulin/rouge/block_entry.py
```diff
@@ -286,6 +293,19 @@ def write(self, fp, offset):
...
+                # create subfolders if needed
+                remote_path_and_name = os.path.split(remote)
+                if remote_path_and_name[0]:
+                    # mmd can't create chain of subfolders
+                    # so we create folders step-by-step
+                    current_path = ""
+                    for folder in remote_path_and_name[0].split('/'):
```
lorc, line 308: Why not `os.path.split`?

moulin/rouge/ext_utils.py
```diff
@@ -79,3 +79,14 @@ def mcopy(img: BinaryIO, file: str, name: str):
...
+
+def mmd(img: BinaryIO, name: str):
+    "Create a directory inside a vfat image"
+
+    # If specified folder already exists, then mmd asks
+    # user for possible action and returns error code.
+    # To avoid such behavior we us '-Ds' as 'skip' clash_option
+    # and ignore error code.
+    args = ["mmd", "-Ds",  "-i", img.name, "::" + name]
+    subprocess.run(args)
```
lorc, line 92: `_run_cmd`

moulin/rouge/block_entry.py
```diff
@@ -271,6 +271,11 @@ def write(self, fp, offset):
             self._complete_init()
         with NamedTemporaryFile() as tempf, TemporaryDirectory() as tempd:
             for remote, local, _ in self._files:
+                # user can specify destination folder from root
+                # and we need to remove very first '/' for correct
+                # work of os.path.join
+                if remote[0] == '/':
```
lorc, line 277: `remote.starts_with('/')` is more pythonic

moulin/rouge/block_entry.py
```diff
@@ -271,6 +271,11 @@ def write(self, fp, offset):
...
+                    remote = remote[1::]
```
lorc, line 278: Second `:` is not needed.

lorc (commented):
I'm not sure about "rouge: reduce flooding of output for user"

Did you tested this with a big images? Like full sd card with android? Is there enough output, so user will be sure that rouge is working and not hung up?

moulin/rouge/block_entry.py
rshym: Thanks, fixed.

moulin/rouge/block_entry.py
rshym: Removed.

moulin/rouge/block_entry.py
rshym: Unfortunately `os.path.split` doesn't provide required functionality.
The target path may contain more than one folder. And `mmd` can create only one folder at a time, it can't work as `os.makedirs` that I used for ext4 partitions.
So I need to get the list of all subfolders, to give them to `mmd` one by one.

rshym:
> I'm not sure about "rouge: reduce flooding of output for user"
> 
> Did you tested this with a big images? Like full sd card with android? Is there enough output, so user will be sure that rouge is working and not hung up?

You are right, I forgot about this case. It will be a really bad experience to see no changes for 10+ minutes.
Commit is removed.

moulin/rouge/ext_utils.py
lorc: Generally, it is not a good idea to ignore errors. Yes, we can expect that subsequent `mcopy` will fail, but nevertheless...

I want to propose third approach: build list of all directories/subdirectories beforehand. Scan all names in `self._files`, build list that includes all directories, remove duplicates, sort it and feed it to one `mmd` call.

rshym: Commit is reworked according to the proposal.

Now for this yaml:
```
files:
  "/zxc/qwe/asd/atf-uboot.ub": "atf-uboot.ub"
  "zxc/qwe/asd/boot.scr": "boot_xen.scr"
  "Image": "Image"
  "/as/system.dtb": "system.dtb"
  "/xc/u-boot.bin": "u-boot.bin"
  "uEnv.txt": "uEnv.txt"
```

We see in the output:
```
[INFO] Running mmd -i /tmp/tmp4jhaa2u9 ::/zxc ::/zxc/qwe ::/zxc/qwe/asd ::/as ::/xc
```

rshym:
Two last commits are reworked.
Tested for `vfat` and `ext4` with the following test set, including different target notations:
```
files:
  "/zxc/qwe/asd/atf-uboot.ub": "atf-uboot.ub"
  "zxc/qwe/asd/boot.scr": "boot_xen.scr"
  "/Image": "Image"
  "/as/system.dtb": "system.dtb"
  "uEnv.txt": "uEnv.txt"
```

Tested with:
- Nested folders with root.
- Nested folders without root.
- Root folder only.
- One level folder.
- No folder at all.

moulin/rouge/block_entry.py
```diff
@@ -285,6 +291,36 @@ def write(self, fp, offset):
...
+                # in incremental way, like `mmd /a /a/s /a/s/d`.
+                # Taking into account that we work with an image file,
+                # we need to prefix each folder with "::".
+                # So here we split each folder into slices,
+                # and re-assemble slices in incremental way.
+                list_for_mmd = list()
+                # remove duplicates by transforming the list of strings to the set of strings
+                for folder in set(list_of_folders):
+                    current_path = ""
+                    for path_slice in folder.split('/'):
```
lorc, line 321: I'm afraid you missed use case when there are files on different levels. Something like this:

```
files:
  "/zxc/atf-uboot.ub": "atf-uboot.ub"
  "zxc/qwe/asd/boot.scr": "boot_xen.scr"
  "/zxc/qwe/atf-uboot.ub": "atf-uboot.ub"
```

Could you please check?

rshym: I missed this case, it will fail.
I will update PR with a proper fix for this case.

lorc: No need in `list_of_folder`. You can build `list_for_mmd` in the very first loop, using the code you wrote there.

Then just do `list_for_mmd = sorted(list(set(list_for_mmd)))` and this should produce correct list in correct order.

rshym: > use case when there are files on different levels

I have fixed the commit and successfully tested it. My previous comment with the list of tested cases was updated accordingly. I have moved `set()` to the end of all manipulations, just before calling `mmd`.

Regarding "No need in `list_of_folder`". Unfortunately, I do not see a possibility to reuse the same list, because in this case, I need to deeply rework the list. Each item from `list_of_folders` may generate more then one item for `list_for_mmd`. Please see below both lists after processing:

```
list_of_folders: ['zxc/qwe/asd', 'zxc/qwe', 'as', 'xc']
list_for_mmd:    ['::/zxc', '::/zxc/qwe', '::/zxc/qwe/asd', '::/zxc', '::/zxc/qwe', '::/as', '::/xc']
```

rshym:
Reworked to fix the issue with files on different levels of nested folders.
The list of tested cases is updated in the previously posted comment.

lorc:
Looks good. Thank you.

rshym:
RB is added.
Ready to be merged.

PR #57, opened 2022-10-04
Fix repo fetcher, yaml_wrapper and rework as_str

rshym:
This PR contains two fixing commits:
- yaml_wrapper: fix copy-paste errors in yaml_wrapper
- fetchers: fix default value for depth

The third commit is proposed for discussion because it changes the behavior of as_str(). My points are explained in the commit message and as comments inside the code.
- yaml_wrapper: allow as_str() to accept any data type

### DIFF
--- a/moulin/yaml_wrapper.py
+++ b/moulin/yaml_wrapper.py
@@ -38,21 +38,21 @@ class _YamlDefaultValue:
     def as_str(self) -> str:
         "Get the string value"
         if not isinstance(self._val, str):
-            raise TypeError("Expected boolean value")
+            raise TypeError("Expected string value")
         return self._val
 
     @property
     def as_int(self) -> int:
         "Get the integer value"
         if not isinstance(self._val, int):
-            raise TypeError("Expected boolean value")
+            raise TypeError("Expected int value")
         return self._val
 
     @property
     def as_float(self) -> float:
         "Get the floating point value"
         if not isinstance(self._val, int):
-            raise TypeError("Expected boolean value")
+            raise TypeError("Expected float value")
         return self._val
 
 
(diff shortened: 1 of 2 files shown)


moulin/yaml_wrapper.py
```diff
@@ -80,9 +80,18 @@ def as_bool(self) -> bool:
...
+        # have fields like 'rev' for 'git' or 'repo' fetcher. That filed always
+        # contains string. And some specific repository may have branch/tag `2.15`.
+        # To handle such case we have two options:
+        #  - obligate a user to use quotes, like "1.45". "1234" or "True";
+        #  - return string presentation of value.
+        # The first option is data-centric and strict, the second one is more user friendly.
+        # Taking into account that strings in YAML are not required to be
+        # quoted if they do not contain some special symbols, we may
+        # consider anything as a string for "obviously" string fields.
+        return str(self._val)
```
lorc, line 94: I'm fine with your intention, but this approach is incorrect.  You are relying on Python to convert scalar back to string, which can result in different value. For example, boolean values in YAML are `true` and `false`. But your code will return Python's `True` and `False`. Same for `null`/`None`.

It is gets even worse with numbers. For example, consider value `0xC`, which will be converted to "12" by your code. `12537e3` which looks like legit git revision would produce "12537000" because it is actually a float number in exponential form.  Or some fancy version like `0.1636` can produce string "0.16359999999" due to float storage limitations.

I skimmed through PyYAML code and didn't found any easy way to access raw values, sadly. But I may something, so you can take a look there too.

lorc (commented):
For two first patches:

As for the last one, as I said in comment, I'm fine with idea, but you need more robust approach.

moulin/yaml_wrapper.py
rshym: Thanks for pointing out these use cases. Looks like it's easier just to use quotes for string.
I will remove the last commit.

Issue #54, opened 2022-09-26
moulin: support network URL for yaml


rshym:
For now, moulin supports only local files as build description.
It could be very convenient to specify `build.yaml` by URL.

rshym:
Need to discuss which type of URL we should support.
List possible cases:
1. raw content on github `https://raw.githubusercontent.com/xen-troops/meta-xt-prod-devel-rcar/master/prod-devel-rcar.yaml`
2. github page `https://github.com/xen-troops/meta-xt-prod-devel-rcar/blob/master/prod-devel-rcar.yaml`
3. A la repo style `-u url -b branch -m name`
4. Something other?

Should we just download build.yaml locally once or re-download it each time?

lorc:
1. For sure
2. Would be good, but is not strictly required
3. This implies dependency on git, I don't think I need this.

> Should we just download build.yaml locally once or re-download it each time?

Download once and refuse to overwrite existing file. I prefer "least surprise" approach.

rshym:
PR is merged.

PR #58, opened 2022-10-05
Add possibility to use URL for build.yaml

rshym:
PR is created to address https://github.com/xen-troops/moulin/issues/54

### DIFF
--- a/moulin/main.py
+++ b/moulin/main.py
@@ -13,6 +13,9 @@ from pathlib import Path
 from typing import Dict, List, Tuple, Any
 import importlib_metadata
 import yaml
+from urllib.parse import urlparse, unquote
+import urllib.request
+import os.path
 
 from packaging.version import Version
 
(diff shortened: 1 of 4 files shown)


rshym:
This PR as addressing https://github.com/xen-troops/moulin/issues/54

moulin/main.py
```diff
@@ -53,6 +56,26 @@ def _prepre_shared_opts(description: str,
     return parser
 
 
+def _download_URL(url: str) -> str:
```
lorc, line 59: well, it not always downloads an URL, as seen in the code.

How about `_get_conf_file`?

Also, probably doc string should be updated accordingly.

rshym:
Rebased to HEAD.
Function `_download_URL` is renamed to `_get_conf_file`, and doc string is updated.
File user-reference.rst is updated with info that file is downloaded only if it doesn't exist locally.
Version is bumped to 0.10.

moulin/main.py
rshym: Done.

moulin/main.py
```diff
@@ -53,6 +56,25 @@ def _prepre_shared_opts(description: str,
     return parser
 
 
+def _get_conf_file(url: str) -> str:
+    "Download file from URL, if it is not present. And return it's local name."
```
lorc, line 60: It is unclear what exactly "is not present". Is it file or URL?

How about "Try to determine config file location. Check if input string is URL and download it if output file is not already present. If input string is not an URL, then assume that this is a local file and return it as is."

moulin/main.py
```diff
@@ -53,6 +56,25 @@ def _prepre_shared_opts(description: str,
     return parser
 
 
+def _get_conf_file(url: str) -> str:
```
lorc, line 59: Input string is not always URL. So parameter name is misleading.

moulin/main.py
rshym: Thanks. I had a problem formulating that. I took your text as is.

moulin/main.py
rshym: `url` replaced with `conf_location`.

rshym:
Comments are addressed.

lorc:

rshym:
RB is added.
Tested, see one of comments above.
Ready to be merged.

PR #60, opened 2022-10-12
Add folders to fill partitions

rshym:
- rouge: use items instead of files
- rouge: minor optimization in block_entry:FileSystem
- rouge: allow directories to be used to fill partition
- bump version to 0.15

### DIFF
--- a/moulin/rouge/block_entry.py
+++ b/moulin/rouge/block_entry.py
@@ -231,20 +231,29 @@ class FileSystem(BlockEntry):
     def __init__(self, node: YamlValue):
         self._node = node
         self._size = 0
-        self._files: List[Tuple[str, str, Mark]] = []
+        self._items: List[Tuple[str, str, Mark]] = []
+
         files_node = self._node.get("files", None)
         if files_node:
+            log.warn("Usage of 'files' is deprecated. Use 'items' please.")
             for remote_node, local_node in cast(YamlValue, files_node).items():
                 remote_name = remote_node
                 local_name = local_node.as_str
-                self._files.append((remote_name, local_name, local_node.mark))
+                self._items.append((remote_name, local_name, local_node.mark))
+
+        items_node = self._node.get("items", None)
+        if items_node:
+            for remote_node, local_node in cast(YamlValue, items_node).items():
+                remote_name = remote_node
+                local_name = local_node.as_str
+                self._items.append((remote_name, local_name, local_node.mark))
 
     def _complete_init(self):
-        for _, local_name, local_mark in self._files:
+        for _, local_name, local_mark in self._items:
             if not os.path.isfile(local_name):
                 raise YAMLProcessingError(f"Can't find file '{local_name}'", local_mark)
 
-        files_size = sum([os.path.getsize(x[1]) for x in self._files]) + 8 * 1024 * 1024
+        files_size = sum([os.path.getsize(x[1]) for x in self._items]) + 8 * 1024 * 1024
         size_node = self._node.get("size", None)
         if size_node:
             self._size = _parse_size(size_node)
@@ -266,7 +275,7 @@ class FileSystem(BlockEntry):
 
     def get_deps(self) -> List[str]:
         "Return list of dependencies needed to build this block"
-        return [f[1] for f in self._files]
+        return [f[1] for f in self._items]
 
 
 class Ext4(FileSystem):
@@ -275,7 +284,7 @@ class Ext4(FileSystem):
         if not self._size:
             self._complete_init()
         with NamedTemporaryFile() as tempf, TemporaryDirectory() as tempd:
-            for remote, local, _ in self._files:
+            for remote, local, _ in self._items:
                 # user can specify destination folder from root
                 # and we need to remove very first '/' for correct
                 # work of os.path.join
@@ -302,7 +311,7 @@ class Vfat(FileSystem):
             ext_utils.mkvfatfs(tempf)
             # scan all remote filenames and collect the list of folders to create
             list_for_mmd = list()
-            for remote, _, _ in self._files:
+            for remote, _, _ in self._items:
                 # remove starting '/' to avoid:
                 # - icluding different forms of same name, like "/zxc" and "zxc"
                 # - adding root folder "/" to list
@@ -329,7 +338,7 @@ class Vfat(FileSystem):
                 # create all destination subfolders at once
                 ext_utils.mmd(tempf, list_for_mmd)
 
-            for remote, local, _ in self._files:
+            for remote, local, _ in self._items:
                 ext_utils.mcopy(tempf, local, remote)
             ext_utils.dd(tempf, fp, offset)
 
--- a/moulin/rouge/block_entry.py
+++ b/moulin/rouge/block_entry.py
@@ -246,10 +246,24 @@ class FileSystem(BlockEntry):
 
     def _complete_init(self):
         for _, local_name, local_mark in self._items:
-            if not os.path.isfile(local_name):
+            if not os.path.isfile(local_name) and not os.path.isdir(local_name):
                 raise YAMLProcessingError(f"Can't find file '{local_name}'", local_mark)
 
-        files_size = sum([os.path.getsize(x[1]) for x in self._items]) + 8 * 1024 * 1024
+        # calculate size of original files and directories
+        files_size = 0
+        for remote_path, local_path, _ in self._items:
+            if os.path.isfile(local_path):
+                files_size += os.path.getsize(local_path)
+            if os.path.isdir(local_path):
+                files_size += os.path.getsize(local_path)
+                for (dirpath, dirnames, filenames) in os.walk(local_path, topdown=True):
+                    for filename in filenames:
+                        # we may have links to location that is incorrect on host, so we use lstat to handle this
+                        files_size += os.lstat(os.path.join(dirpath, filename)).st_size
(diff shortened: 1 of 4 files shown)


rshym:
Testing is ongoing.

rshym:
Tested filling of `vfat` partition by nested folders to destination root, and to nested folders on partition.
Symlinks are not supported for `vfat`.

Tested usage of `files:` and `folders:` descriptors in all three cases:
- `files` only
- `folders` only
- `files` and `folders` at the same time

rshym:
One comment. I'm not sure that we need to separate `files:` and `folders:`. Maybe it's better to unite them into `files` with the possibility to provide folders as well.

moulin/rouge/block_entry.py
```diff
@@ -306,7 +306,7 @@ def write(self, fp, offset):
                     # create destination subfolder
                     os.makedirs(os.path.join(tempd, remote_path_and_name[0]), exist_ok=True)
 
-                shutil.copyfile(local, os.path.join(tempd, remote))
+                shutil.copyfile(local, os.path.join(tempd, remote), follow_symlinks=False)
```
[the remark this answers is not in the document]
rshym: I have tried to find some way to modify absolute symlinks, but I think that it's not a good idea to modify the content of symlink. User may have symlinks that will be correct on target only or intentionally contains an absolute path.
So, it's better to copy symlinks without any modifications.

But I made additional tests and found that the symlink to a directory is ignored. So I made additional changes in the newer version of the commit.

rshym:
@lorc could you please take a look again?

- files replaced with items
- folders are not used anymore
- docs are updated accordingly
- added handling of a symlink to a directory
- fixed Lint's error "C901 'FileSystem._complete_init' is too complex", appeared after adding symlinks
- version is incremented

moulin/rouge/block_entry.py
```diff
@@ -266,7 +275,9 @@ def write(self, fp, offset):
 
     def get_deps(self) -> List[str]:
         "Return list of dependencies needed to build this block"
-        return [f[1] for f in self._files]
+        deps: List[str] = [f[1] for f in self._files]
+        deps.extend(f[1] for f in self._items)
```
lorc, line 310: Looks like you didn't introduced `_items` in this patch.

moulin/rouge/block_entry.py
```diff
@@ -246,9 +249,27 @@ def __init__(self, node: YamlValue):
             for remote_node, local_node in cast(YamlValue, items_node).items():
                 remote_name = remote_node
                 local_name = local_node.as_str
-                self._files.append((remote_name, local_name, local_node.mark))
+                self._items.append((remote_name, local_name, local_node.mark))
 
     def _complete_init(self):
+        # sort items out
+        for remote_path, local_path, yaml_mark in self._items:
```
lorc, line 262: Do we need all this complex logic at  all? Having two separate arrays: `_items` and `_files`, scanning directory to build list of files... 

Can we just have one `_items` array and store both actual files and directories in that array? In case of ext4, we can use `shutil.copytree` and so be it. In case of vfat, we probably need this logic, so we can move it into vfat code.

Also, I just found `shutil.disk_usage` that can help on calculating needed space.

moulin/rouge/block_entry.py
```diff
@@ -257,7 +257,7 @@ def _add_to_files(self, relative_dir: str, name: str, remote_path: str, dirpath:
         log.debug(remote_filename + " <<  " + relative_local_filename)
         self._files.append((remote_filename, os.path.join(dirpath, name), yaml_mark))
 
-    def _complete_init(self):
+    def _generate_list_of_files(self):
```
lorc, line 260: It will be good to move this change into the original commit. Then, there will be no need for this commit at all.

lorc (commented):
Thank for the changes. I have a couple more comments.

rshym:
@lorc could you please take a look again?
All changes are fully reworked and tested with the following yaml-code (vfat tested as well):
```
images:
  test:
    type: gpt
    desc: "test image"
    partitions:
      test:
        gpt_type: 21686148-6449-6E6F-744E-656564454649 # BIOS boot partition (kinda...)
        type: ext4
        items:
          "file1": "./file_1"
          "folderA/fileA": "dirA/file_A"
          "folderB": "dirA"
          "/": "dirA/"
          "/folderC/subA": "dirA/"
```

moulin/rouge/block_entry.py
rshym: Unfortunately, `shutil.disk_usage` can be used here, because it returns the usage of space on disk, that hosts specified directory. So I didn't find a better way to calculate the size of the directory than `os.walk`.
Other comments are addressed.

moulin/rouge/block_entry.py
```diff
@@ -246,10 +246,24 @@ def __init__(self, node: YamlValue):
...
+            if os.path.isfile(local_path):
+                files_size += os.path.getsize(local_path)
+            if os.path.isdir(local_path):
+                files_size += os.path.getsize(local_path)
+                for (dirpath, dirnames, filenames) in os.walk(local_path, topdown=True):
+                    for filename in filenames:
+                        # we may have links to location that is incorrect on host, so we use lstat to handle this
+                        files_size += os.lstat(os.path.join(dirpath, filename)).st_size
+                    for dirname in dirnames:
+                        files_size += os.path.getsize(os.path.join(dirpath, dirname))
```
lorc, line 264: This is strange. I believe `os.walk` will enter into the directories. Why do you need to add dir size?

moulin/rouge/block_entry.py
```diff
@@ -291,20 +305,43 @@ def write(self, fp, offset):
...
+        out_list: List[Tuple[str, str, Mark]] = []
+        for remote, local, mark in self._items:
+            if os.path.isfile(local):
+                out_list.append([remote, local, mark])
+            if os.path.isdir(local):
+                for (dirpath, _, filenames) in os.walk(local, topdown=True):
+                    remote_dirpath = dirpath.replace(local, remote, 1)
+                    for filename in filenames:
+                        # we skip symlinks as not supported on vfat
+                        if not os.path.islink(os.path.join(dirpath, filename)):
```
lorc, line 331: Could you print some warning if you encounter a link? I believe user needs to know if symlink was skipped.

lorc (commented):
Thank you. I'm generally fine with the changes. There a just couple of minor comments.

moulin/rouge/block_entry.py
rshym: Yes, that's a good idea.
I have added a warning. See below an example of the output:
```
[WARNING] Symlink 'dirA/link_to_file' is skipped.
```

moulin/rouge/block_entry.py
rshym: Sorry, if you will write something that is well-known to you, I just try to make a consistent and full answer.

The initial idea was that in some cases directory may be quite big itself. I have tested that by putting 150 small files into the directory.

For example, the minimum size of an empty directory is 4096 bytes.
If I put 150 small files (465 bytes each) into the directory, then `getsize()` for the directory return 12288 bytes. And after that increase, the size will not shrink back to 4096 even if those files are removed.

Also, `getsize()` for the directory doesn't include the size of contained files.

Let me show the following example, which is my actual test setup:
```
dirA\
  dirB\
    dirXX\
      files
    files
  dirC\
    file
  files...
  one_link_to_directory
```

On each iteration of `os.walk()` we do not actually enter inside subdir, but return the set of variables for that subdir.
As you can see below, for `dirB` we get set of variables:
```
dirpath: dirA/dirB
dirnames: ['dirXX']
filenames: ['file3', 'link_to_fileA']
```

```
[WARNING] ==========
[WARNING] Entry point for directory 'dirA/' size is 12288
[WARNING]   dirpath: dirA/
[WARNING]   dirnames: ['link_to_dir', 'dirB', 'dirC']
[WARNING]   filenames: ['bl2-h3ulcb-4x2g.bin', 'link_to_file', 'file_A', 'link_to_abs']
[WARNING] file 'dirA/bl2-h3ulcb-4x2g.bin' size is 141094
[WARNING] file 'dirA/link_to_file' size is 19
[WARNING] file 'dirA/file_A' size is 465
[WARNING] file 'dirA/link_to_abs' size is 14
[WARNING] dir 'dirA/link_to_dir' size is 12288
[WARNING] dir 'dirA/dirB' size is 12288
[WARNING] dir 'dirA/dirC' size is 12288
[WARNING]   dirpath: dirA/dirB
[WARNING]   dirnames: ['dirXX']
[WARNING]   filenames: ['file3', 'link_to_fileA']
[WARNING] file 'dirA/dirB/file3' size is 141094
[WARNING] file 'dirA/dirB/link_to_fileA' size is 9
[WARNING] dir 'dirA/dirB/dirXX' size is 4096
[WARNING]   dirpath: dirA/dirB/dirXX
[WARNING]   dirnames: []
[WARNING]   filenames: ['file_A_XX', 'file3_XX']
[WARNING] file 'dirA/dirB/dirXX/file_A_XX' size is 465
[WARNING] file 'dirA/dirB/dirXX/file3_XX' size is 141094
[WARNING]   dirpath: dirA/dirC
[WARNING]   dirnames: []
[WARNING]   filenames: ['file_A']
[WARNING] file 'dirA/dirC/file_A' size is 465
```

rshym:
`Ack` is added.
Tested.
Ready to be merged.

PR #62, opened 2022-12-15
zephyr: add `env` option to builder

rshym:
This option allows to provide custom environment variable available during build of Zephyr application.

Example of usage:

```
    builder:
      type: zephyr
      <...>
      env:
        - "MY_ENV_VAR=my_value"
```

Signed-off-by: Ruslan Shymkevych <ruslan_shymkevych@epam.com>

### DIFF
--- a/docs/user-reference.rst
+++ b/docs/user-reference.rst
@@ -572,6 +572,9 @@ to be fetched by `west` fetcher.
     target: samples/synchronization
     target_images:
       - "zephyr/build/zephyr/zephyr.bin"
+    env:
+      - "MY_ENV_VAR=my_value"
+
 
 Mandatory options:
 
@@ -589,6 +592,11 @@ Mandatory options:
 * :code:`target_images` - list of image files that should be generated
   by this builder. For standard build, it is expected to be
   "zephyr/build/zephyr/zephyr.bin"
+
+Optional parameters:
+
+* :code:`env` - list of additional environment variables that should
+  be exported before calling :code:`west build`.
 
 Please note that this builder uses :code:`--pristine=auto` command-line option.
 
--- a/moulin/builders/zephyr.py
+++ b/moulin/builders/zephyr.py
@@ -28,7 +28,7 @@ def gen_build_rules(generator: ninja_syntax.Writer):
         construct_fetcher_dep_cmd(),
         "cd $build_dir/zephyr",
         "source zephyr-env.sh",
-        "west build -p auto -b $board $target",
+        "$env west build -p auto -b $board $target",
     ])
     generator.rule("zephyr_build",
                    command=f'bash -c "{cmd}"',
@@ -54,11 +54,20 @@ class ZephyrBuilder:
 
     def gen_build(self):
         """Generate Ninja rules to build Zephyr"""
+
+        env_node = self.conf.get("env", None)
+        if env_node:
+            env_values = [x.as_str for x in env_node]
+        else:
+            env_values = []
+        env = " ".join(env_values)
+
         variables = {
             "name": self.name,
             "build_dir": self.build_dir,
             "board": self.conf["board"].as_str,
             "target": self.conf["target"].as_str,
+            "env": env,
         }
         targets = self.get_targets()
         deps = list(self.src_stamps)


docs/user-reference.rst
```diff
@@ -572,6 +572,9 @@ to be fetched by `west` fetcher.
     target: samples/synchronization
     target_images:
       - "zephyr/build/zephyr/zephyr.bin"
+    additional_env_vars:
+      - [MY_ENV_VAR, "my_value"]
```
lorc, line 576: Yocto builder uses this format because there can identical entries:

`SOME_VAR_append="aaa"`
`SOME_VAR_append="bbb"`

This is not the case for environment variables. So maybe dict will be better in this case?

moulin/builders/zephyr.py
```diff
@@ -28,6 +28,7 @@ def gen_build_rules(generator: ninja_syntax.Writer):
         construct_fetcher_dep_cmd(),
         "cd $build_dir/zephyr",
         "source zephyr-env.sh",
+        "$additional_env_vars",
```
lorc, line 31: Not sure how this should work...

Vars will be joined with `&&`, so resulting command will look like this:

`cd $build_dir/zephyr && source zephyr-env.sh && VAR1=1 VAR2=2 && west build -p auto -b $board $target`

WIll it cause shell to carry variable values past `&&`?


I'd prefer it to look like this:

`cd $build_dir/zephyr && source zephyr-env.sh && VAR1=1 VAR2=2 west build -p auto -b $board $target`

In this case you will not need hack with `true` in case of absent variables

docs/user-reference.rst
```diff
@@ -572,6 +572,9 @@ to be fetched by `west` fetcher.
     target: samples/synchronization
     target_images:
       - "zephyr/build/zephyr/zephyr.bin"
+    additional_env_vars:
```
lorc, line 575: Isn't `additional_env_vars` to mouthull? Why not just `env` as we use in Android builder? It will be more consistent.

moulin/builders/zephyr.py
```diff
@@ -63,6 +85,15 @@ def gen_build(self):
...
         deps = list(self.src_stamps)
 
+        # Additional environment variables
+        additional_env_vars_node = self.conf.get("additional_env_vars", None)
+        if additional_env_vars_node:
+            self.additional_env_vars = _flatten_env_vars(additional_env_vars_node)
+            variables = dict(variables, additional_env_vars=self.additional_env_vars)
+        else:
+            # if additional_env_vars is not set, we need to set it to NOP command
+            variables = dict(variables, additional_env_vars="true")
```
lorc, line 95: With my proposed change above, this will not be needed

moulin/builders/zephyr.py
rshym: Done.

docs/user-reference.rst
rshym: Renamed to `env`.

moulin/builders/zephyr.py
rshym: Removed.

docs/user-reference.rst
rshym: Thanks for your comment below I decided to use the same approach as in android_kernel:
```
env:
  - "MY_VAR=my_val"
```
This made the commit much simpler.

rshym:
All comments are addressed.
Tested with `ninja -v` to check the command line for different use cases (one var, few vars, no vars).

moulin/builders/zephyr.py
```diff
@@ -54,11 +54,20 @@ def __init__(self, conf: YamlValue, name: str, build_dir: str, src_stamps: List[
 
     def gen_build(self):
         """Generate Ninja rules to build Zephyr"""
+
+        env_node = self.conf.get("env", None)
```
lorc, line 58: `env_node = self.conf.get("env", [])` and remove `if` below. Probably, you can even make it one-liner.

lorc (commented):
With that last change:

moulin/builders/zephyr.py
rshym: Unfortunately, this can't be done, because
```
'_YamlDefaultValue' object is not iterable
```

rshym:
Commit is renamed to comply with the used practice (start with 'topic:').

moulin/builders/zephyr.py
rshym: Implementation of the requested change requires one of the following:
- implementation of an iterator for _YamlDefaultValue;
- change of interface of `conf.get` to `... -> YamlValue`, so returned object will always be iterable
- adding a new method that will be able always to return an iterable object

In all cases, this requires additional changes and discussion of modification of the interface of yaml_wrapper.

So, if you agree, I will leave the code as is.

lorc: Ah, I see. Yes, it is easier to leave it as is for now.

lorc (approved):
You can add my r-b tag

rshym:
'R-B' is added.
Tested.
Ready to be merged.

PR #63, opened 2022-12-18
zephyr: add ext_files option to builder

rshym:
This option allows to collect external files that are required for the build.
Example of corresponding yaml:

```
  builder:
    type: zephyr
    ...
    ext_files:
      "dest_file_1": "path/to/external/file_1"
      "abc/file_BB": "path/to/external/file_2"
```
Destination directories are created if specified.
File may be renamed to desired destination name.

Signed-off-by: Ruslan Shymkevych <ruslan_shymkevych@epam.com>

### DIFF
--- a/docs/user-reference.rst
+++ b/docs/user-reference.rst
@@ -574,6 +574,9 @@ to be fetched by `west` fetcher.
       - "zephyr/build/zephyr/zephyr.bin"
     env:
       - "MY_ENV_VAR=my_value"
+    ext_files:
+      "file_1": "path/to/external/file_1"
+      "abc/file_BB": "path/to/external/file_2"
 
 
 Mandatory options:
@@ -597,6 +600,13 @@ Optional parameters:
 
 * :code:`env` - list of additional environment variables that should
   be exported before calling :code:`west build`.
+
+* :code:`ext_files` - defines :code:`destination_name:path_to_source_file`
+  mapping of external files that need to be collected before calling
+  :code:`west build`. The destination is specified relative to zephyr's
+  :code:`build-dir`, and can include subdirectories. If the destination
+  path is specified then it will be created. You can rename the source
+  file by providing its new name in the destination part.
 
 Please note that this builder uses :code:`--pristine=auto` command-line option.
 
--- a/moulin/builders/zephyr.py
+++ b/moulin/builders/zephyr.py
@@ -28,10 +28,7 @@ def gen_build_rules(generator: ninja_syntax.Writer):
         construct_fetcher_dep_cmd(),
         "cd $build_dir",
         "$pull_ext_sources",
-        # Variable 'pull_ext_sources' always has "cd zephyr" appended,
-        # to avoid '&&  &&' if 'ext_files' are not provided.
-        # Other possible solution - to use 'true' command when 'ext_files' are absent.
-        # "cd zephyr",
+        "cd zephyr",
         "source zephyr-env.sh",
         "$env west build -p auto -b $board $target",
     ])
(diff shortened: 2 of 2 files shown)


rshym:
Tested with yaml containg:
```
      ext_files:
        # destination paths are relative to %{build-dir}
        "xenutils/include/domain_bins/domd_img.bin": "%{YOCTOS_WORK_DIR}/%{DOMD_BUILD_DIR}/tmp/deploy/images/%{MACHINE}/Image"
        "xenutils/include/domain_bins/domd.dtb": "%{YOCTOS_WORK_DIR}/%{DOMD_BUILD_DIR}/tmp/deploy/images/%{MACHINE}/%{XT_DOMD_DTB_NAME}"
        "xenutils/include/domain_bins/domu_img.bin": "%{YOCTOS_WORK_DIR}/%{DOMD_BUILD_DIR}/tmp/deploy/images/%{MACHINE}/Image"
        "xenutils/include/domain_bins/domu.dtb": "%{YOCTOS_WORK_DIR}/%{DOMD_BUILD_DIR}/tmp/deploy/images/%{MACHINE}/%{XT_DOMD_DTB_NAME}"
```
Also, tested:
- without destination directories
- with absence of `ext_files` (that why I have added `|| true")

docs/user-reference.rst
```diff
@@ -598,6 +601,13 @@ Optional parameters:
 * :code:`env` - list of additional environment variables that should
   be exported before calling :code:`west build`.
 
+* :code:`ext_files` - list of external files that need to be collected
```
lorc, line 604: Well, this is not a list. It is mapping. YAML uses a different syntax for lists

docs/user-reference.rst
```diff
@@ -598,6 +601,13 @@ Optional parameters:
...
+  before calling :code:`west build`. These are pairs of
+  `"destination name": "path to source file"`. Destination is specified
+  relative to zephyr's `build-dir`, and can contain path. If destination
```
lorc, line 607: s/"can contain path"/"can include subdirectories" maybe?

moulin/builders/zephyr.py
```diff
@@ -26,7 +28,10 @@ def gen_build_rules(generator: ninja_syntax.Writer):
     cmd = " && ".join([
         # Generate fetcher dependency file
         construct_fetcher_dep_cmd(),
-        "cd $build_dir/zephyr",
+        "cd $build_dir",
+        # script is created only is 'ext_files' are specified, so we need '|| true'
+        "source " + SCRIPT_NAME_TO_PULL_BINS + " || true",
```
lorc, line 33: You could use string interpolation (aka f-string) there, but this is not the biggest problem.

I don't think we need a separate script there at all. We may just use variable there. It will be a bit less awful. Just replace this whole line with `($pull_ext_sources)` and then provide variable `pull_ext_sources` in `gen_build` with contents of your `SCRIPT_NAME_TO_PULL_BINS`

moulin/builders/zephyr.py
```diff
@@ -55,6 +60,27 @@ def __init__(self, conf: YamlValue, name: str, build_dir: str, src_stamps: List[
     def gen_build(self):
         """Generate Ninja rules to build Zephyr"""
 
+        ext_files_node = self.conf.get("ext_files", None)
+        if ext_files_node:
+            files: List[Tuple[str, str]] = []
+            for dest_name, src_node in cast(YamlValue, ext_files_node).items():
+                src_name = os.path.relpath(src_node.as_str, self.build_dir)
```
lorc, line 67: I believe we need to add each SRC to dependencies list, so Ninja will trigger rebuild if external file is changed.

moulin/builders/zephyr.py
rshym: That was my initial idea, but I'm afraid that we may reach some limits if we will have 10+ files with long paths.
So I decided to use the external script which may handle any amount of files.
But I may be wrong regarding command-line limits.

lorc: AFAI, command line limit in Linux is about 64 kbytes, which should be enough in all cases except Android builds :)

docs/user-reference.rst
rshym: Thanks. This part if reworded.

docs/user-reference.rst
rshym: Done.

moulin/builders/zephyr.py
rshym: Reworked. The additional script is not used. All commands are put into the command line. Please see the comment in the code regarding the handling of empty `ext_files`.

moulin/builders/zephyr.py
rshym: Reworked. Tested. Build of DomD was initiated.

moulin/builders/zephyr.py
```diff
@@ -55,6 +60,22 @@ def __init__(self, conf: YamlValue, name: str, build_dir: str, src_stamps: List[
...
+                ext_deps.append(src_node.as_str)
+                # create a-la "script" to copy external files to build-dir before start of build
+                # do we need to create destination subfolders?
+                dst_path_and_name = os.path.split(dest_name)
+                if dst_path_and_name[0]:
+                    script.append(f"mkdir -p {dst_path_and_name[0]}")
+                script.append(f"cp {src_name} {dest_name}")
+        script.append("cd zephyr")
```
lorc, line 76: Why do you need `cp zephyr` and end of the script?

moulin/builders/zephyr.py
```diff
@@ -26,7 +26,12 @@ def gen_build_rules(generator: ninja_syntax.Writer):
     cmd = " && ".join([
         # Generate fetcher dependency file
         construct_fetcher_dep_cmd(),
-        "cd $build_dir/zephyr",
+        "cd $build_dir",
+        "$pull_ext_sources",
+        # Variable 'pull_ext_sources' always has "cd zephyr" appended,
```
lorc, line 31: This is wrong. You are mixing two functionalities together.  Better leave `true` in the `pull_ext_sources`.

lorc (changes requested):

moulin/builders/zephyr.py
rshym: Comments are reworked.

moulin/builders/zephyr.py
rshym: This approach was reworked. Part of the "outer" build script is not mixed with "pull external files" anymore.

moulin/builders/zephyr.py
rshym: This approach was reworked.

PR #65, opened 2023-01-27
Update syntax for EXTERNALSRC_pn-{key}

otyshchenko1:
Since the Honister release the "_" character needs to be replaced with ":".

In particular, wrong syntax in that variable(s) being written to moulin.conf leads to the domd.bb(domu.dd):do_install tasks failure.

Signed-off-by: Oleksandr Tyshchenko <oleksandr_tyshchenko@epam.com>

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,7 +21,7 @@ copyright = '2021-2022, EPAM Systems'
 author = 'EPAM Systems'
 
 # The full version, including alpha/beta/rc tags
-release = 'v0.11'
+release = 'v0.12'
 
 # -- General configuration ---------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


lorc:
Good, thank you.

Two comments:

1.  we need to be sure that older versions of Yocto/Poky support this new syntax.
2. please bump up Moulin version, so we can use proper `min_var` value in YAML files

otyshchenko1:
> Good, thank you.
> 
> Two comments:
> 
> 1. we need to be sure that older versions of Yocto/Poky support this new syntax.

sure

> 2. please bump up Moulin version, so we can use proper `min_var` value in YAML files

will do

rshym:
Tested with yocto dunfell. Works as expected.

lorc:

PR #68, opened 2023-03-13
zephyr builder:  additional_deps

dsemenets:


lorc:
Thanks, I'm fine with the patch. It is good that you updated documentation, but you forget to add deeper explanation in "Optional parameters:" section. Could you please add it? I believe, you can copy text from Yocto's section, just update it accordingly.

lorc:
Thank you

lorc:
Okay, I can see that you want to provide paths with external files to Zephyr, but how Zephyr will handle this? Is there an example?

dsemenets:
For now Zephyr doesn't handle this. This option will be used in future I guess.

lorc:
Then, I'm against this patch. We need this feature, yes, but it may come that we need to implement this in a some other way. E.g. by generating some file YAML for `west` instead of providing environment variables.

moulin/builders/zephyr.py
```diff
@@ -78,12 +78,24 @@ def gen_build(self):
...
         env = " ".join(env_values)
 
+        shields_node = self.conf.get("shields", None)
+        if shields_node:
+            shields_vals = [x.as_str for x in shields_node]
+        else:
+            shields_vals = None
+
+        if shields_vals:
+            shields = "-DSHIELD=\\\""+" ".join(shields_vals)+"\\\""
```
lorc, line 88: This is not pythonic.

1. Have you tested case where you have more than one shield? Do Zephyr build system handles `-DSHIELD="shield1 shield2"`? 

2. You might rewrite this in the following way: `shields = f'-DSHIELD="{" ".join(shields_vals)}"'`

moulin/builders/zephyr.py
```diff
@@ -60,6 +75,7 @@ def gen_build(self):
             env_values = [x.as_str for x in env_node]
         else:
             env_values = []
+        env_values.extend([v+"=\""+val+"\"" for v, val in self._get_external_src()])
```
lorc, line 78: Python allows both `'` and `"` to define string literals.  You can go away from escaping `"`. Also, python supports string interpolation: 

`env_values.extend([f'{v}="{val}"' for v, val in self._get_external_src()])`

lorc:
Okay, I'm fine with the most changes, but please remove `builders: zephyr: Add external_src option to zephyr builder`. As I said earlier, we need working Zephyr counterpart before implementing something in Moulin.

moulin/builders/zephyr.py
```diff
@@ -62,12 +62,24 @@ def gen_build(self):
             env_values = []
         env = " ".join(env_values)
 
+        shields_node = self.conf.get("shields", None)
+        if shields_node:
```
lorc, line 66: May I ask you to fold the next `if` into this one?  It will be much cleaner:

```
        if shields_node:
            shields_vals = [x.as_str for x in shields_node]
            shields = f'-DSHIELD=\\"{" ".join(shields_vals)}\\"'
	else:
            shields = ""
```

lorc (commented):
With the last comment fixed:

lorc:
Thank you!

PR #69, opened 2023-04-04
builders: archive: introduce tar's base dir

rshym:
This commit solves an issue that archives created by the archive builder
contain paths to files relative to build-dir. As result, paths may
contain a long list of build-related directories, and in some cases, this
is not desirable.
A possible solution is to specify desired starting point as build dir,
but this will put the resulting archive into the same directory.
So, the best way to get an archive in the top directory and get rid of
intermediate directories - is to specify the base directory for `tar`
through the `-C` option.

That's why this commit introduces the `base_dir` field for the archive
builder. The `base_dir` is an optional field, and it has '.' value if
not provided.

### DIFF
--- a/docs/user-reference.rst
+++ b/docs/user-reference.rst
@@ -529,7 +529,7 @@ Optional parameters:
 archive builder
 ^^^^^^^^^^^^^^^
 
-Archive builder does is intended to create archive from other components.
+Archive builder is intended to create an archive from other components.
 It can be used to gather build artifacts, for example. This builder
 uses `tar` to create archive files. Archives can be optionally compressed
 as, `tar` is invoked with `--auto-compress` option.
@@ -537,7 +537,7 @@ as, `tar` is invoked with `--auto-compress` option.
 .. code-block:: yaml
 
   builder:
-    type: archive        # Should be 'artchive'
+    type: archive        # Should be 'archive'
     name: "artifacts.tar.bz2"
     items:
       - "yocto/build/tmp/deploy/images/generic-armv8-xt/Image"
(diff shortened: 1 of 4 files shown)


docs/user-reference.rst
```diff
@@ -551,6 +552,13 @@ Mandatory options:
 * :code:`name` - Name of an archive file. Add a suffix like `tar.bz2` to
   make `tar` compress archive with desired compressing algorithm.
 
+* :code:`work_dir` - `tar`'s work directory. Optional parameter.
```
lorc, line 555: I believe `base_dir` will be better name. "Work directory" is a place where some work is done. Which is not quite true in this case.

docs/user-reference.rst
```diff
@@ -551,6 +552,13 @@ Mandatory options:
...
+  The default value is "." if not specified. This is passed to `tar` as
+  `-C` option. As result, the final archive will contain paths relative
+  to :code:`work_dir`. In the example above, `artifacts.tar.bz2` will
+  contain files `generic-armv8-xt/Image` and `generic-armv8-xt/uInitramfs`.
```
lorc, line 559: I am wondering what will happen to items that are not in the `work_dir`/`base_dir` subtree. For example, what if we'll have `yocto/build-dom/some-file` in `items`? 

I believe it is better to declare that all items paths should be relative to `work_dir`/`base_dir`.

docs/user-reference.rst
rshym: Renamed.

docs/user-reference.rst
rshym: You are right. I have checked with the following config:
```
    builder:
      type: archive
      name: "zxc.tar.bz2"
      base_dir: "a/b/c"
      items:
        - "a/b/c/d/fileA.txt"
        - "a/b/c/d/fileB.txt"
        - "a/bb/cc/build.ninja"
```
and got results:
```
$ ninja -v pack-ipl
[1/1] tar --sparse --dereference --create --auto-compress -C a/b/c --file=zxc.tar.bz2 d/fileA.txt d/fileB.txt ../../bb/cc/build.ninja
tar: Removing leading `../../' from member names
tar: Removing leading `../../' from hard link targets


$ tar -tf zxc.tar.bz2 
d/fileA.txt
d/fileB.txt
bb/cc/build.ninja
```

So I have added an additional note to the documentation.

rshym:
I have fixed the misprint in the text.
Ready for a new review.

rshym:
Sources and documentation are reworked. Items now should be specified relative to `base_dir`.

lorc:

rshym:
`Ack` is added.
Tested (see the comment above).
Ready to be merged.

PR #76, opened 2023-06-05
raise exception when have ambiguousness of revisions

iusyk:
yaml file may have some ambiguousness of revisions, for example .....
domd:
    sources:
      - type: git url: "https://github.com/good/meta-good" .....
--ENABLE_SOMETHING
      default: no
      ......
         components:
            domd:
              sources:
                - type: git url: "https://github.com/good/meta-good" rev: main Here we have one repository but with the two different revisions. At this moment, moulin checks if url has been fetched and skip all the overrides. It means that build may have unexpected result. The patch check if the repos exists and if the revisions are the same. If revisions are not the same -  throws the exception.

### DIFF
--- a/moulin/build_generator.py
+++ b/moulin/build_generator.py
@@ -5,6 +5,7 @@ This module takes processed YAML config tree (with variables expanded
 and parameters applied) and produces Ninja build file
 """
 
+from os import remove
 import os.path
 import sys
 from typing import Optional, List
@@ -50,7 +51,16 @@ def generate_build(conf: MoulinConfiguration,
                 source_type = source["type"].as_str
                 fetcher_module = fetcher_modules[source_type]
                 fetcher = fetcher_module.get_fetcher(source, build_dir, generator)
-                fetcher_stamps = fetcher.gen_fetch()
+                try:
+                    fetcher_stamps = fetcher.gen_fetch()
+                except Exception:
+                    generator.close()
+                    # file is not usable
+                    remove(ninja_build_fname)
+                    # Raise an exception to delegate the decision
+                    # at the higher lever
+                    raise
+
                 if isinstance(fetcher_stamps, list):
                     source_stamps.extend(fetcher_stamps)
                 else:
--- a/moulin/fetchers/git.py
+++ b/moulin/fetchers/git.py
@@ -44,7 +44,7 @@ def _guess_dirname(url: str):
     return url.split("/")[-1]
 
 
-_SEEN_REPOS = []
+_SEEN_REPOS_REV = {}
 
 
 class GitFetcher:
@@ -64,10 +64,15 @@ class GitFetcher:
         checkout_stamp = create_stamp_name(self.build_dir, self.url, "checkout")
 
         # Do not checkout repos for the second time
-        if checkout_stamp in _SEEN_REPOS:
-            return checkout_stamp
+        if checkout_stamp in _SEEN_REPOS_REV:
+            if self.git_rev != _SEEN_REPOS_REV[checkout_stamp]:
+                raise Exception(f"ERROR: Repository {self.url} has two\
+ revisions '{self.git_rev}' and '{_SEEN_REPOS_REV[checkout_stamp]}'")
+            else:
+                return checkout_stamp
 
-        _SEEN_REPOS.append(checkout_stamp)
+        _SEEN_REPOS_REV[checkout_stamp] = self.git_rev
+
         self.generator.build(clone_target,
                              "git_clone",
                              variables={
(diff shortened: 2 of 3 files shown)


moulin/fetchers/git.py
```diff
@@ -64,10 +64,18 @@ def gen_fetch(self):
         checkout_stamp = create_stamp_name(self.build_dir, self.url, "checkout")
 
         # Do not checkout repos for the second time
-        if checkout_stamp in _SEEN_REPOS:
-            return checkout_stamp
+        if (checkout_stamp in _SEEN_REPOS_REV and
+                self.git_rev != _SEEN_REPOS_REV.get(checkout_stamp)):
```
lorc, line 68: 1. this is not C, so no need for parentheses in the `if` condition
2. There is no need to use `get`, you can use `[]` syntax. 
3. You missed the crucial point of returning stamp if it is already present. Now mulin will generate multiple "build" entries for each git repo which is bad

moulin/fetchers/git.py
```diff
@@ -64,10 +64,18 @@ def gen_fetch(self):
...
+            raise Exception("ERROR: Repository {} has the two revisions, \
+                            {} and {}, fix the ambiguousness."
+                            .format(
```
lorc, line 71: Please use string interpolation (aka f-stings) instead of `format`.

We want project source code to be consistent in used techniques.

lorc (changes requested):

moulin/fetchers/git.py
iusyk: 1. fixed
2. fixed
3. generator is closed, file is removed and ruining is stopped

moulin/fetchers/git.py
iusyk: fixed

moulin/fetchers/git.py
lorc: 3. Whoa. You don't need to do all of this. Please remove all those changes, they are not necessary. 

My point was that original code issued `return checkout_stamp`. You removed this part, so now every call for the the same git repo will create a new `build` entry in the `build.ninja`.

iusyk: No, if everything is ok - we have 'return checkout_stamp' at the end of the function, if we have the same repo but different revisions - tool must be stopped. Bit yes, one case has not been considered correctly - when we have the same revisions, in this case call 'return checkout_stamp'.

lorc (changes requested):
Let's stick only to git fetcher changes for now.

moulin/build_generator.py
```diff
@@ -5,6 +5,7 @@
 and parameters applied) and produces Ninja build file
 """
 
+from os import remove
```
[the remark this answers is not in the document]
iusyk: then, we have not too much to do. 
1. leave as is (what is wrong) and close the PR
2. inform (via logging) user about the problem.
Raising the exception without the catching it does not make any sense.
The choice ?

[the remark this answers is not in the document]
lorc: > leave as is (what is wrong) and close the PR

Why? You made a valid change to `git.py`. I want to merge it.

> Raising the exception without the catching it does not make any sense.

Again, why? Python will catch it for us and write exception message with the trace log.

Look, I am not against more pretty error reporting, but it should be done properly:
1. In a separate commit (do not forget about atomic commits, please)
2. In `moulin_entry()` function, as it is main entry point. `moulin.py` is used only during local debugging. It is not installed and not used with `pip install`
3. Print exception backtrace, so user can trace  error to the correct place in the code. This is very important in case of YAML parsing exceptions, because it is not immediately obvious which code triggered an exception.
4.  Do the same for `rouge`
5. Use `log.error()`, not `print`
6. Cover all errors, not only `fetcher.gen_fetch()` call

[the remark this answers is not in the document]
iusyk: well, in the case if we relay on python - build.ninja will be generated, it is better to avoid it.
Backtrace - i think this is an error for user , not for developer to fix the bug, but backtrace is useful i agree
i review rouge as well to see what to do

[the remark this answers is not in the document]
lorc: As I said, backtrace can be handy for users to debug errors in YAML files, because backtrace allows user to see which component triggered an error.

lorc:
Can we limit this PR only to "one git repo, different revisions" bug fix? It is almost ready, you just need to remove unnecessary changes. We can merge it pronto.

You can open additional PR with enhanced error handling. I am envisioning that there will be much more discussions, so it will take much more time before it could be merged.

PR #79, opened 2023-06-16
rouge: create vfat filesystem with proper sector size

sa-kib:
Fix regression from custom sector size feature -- vfat FS has to be created with correct sector size provided, otherwise it can't be mounted on running target system.

### DIFF
--- a/moulin/rouge/block_entry.py
+++ b/moulin/rouge/block_entry.py
@@ -47,7 +47,7 @@ class GPTPartition(NamedTuple):
 class GPT(BlockEntry):
     "Represents GUID Partition Table"
 
-    def __init__(self, node: YamlValue):
+    def __init__(self, node: YamlValue, **kwargs):
         self._partitions: List[GPTPartition] = []
         self._size: int = 0
         self._sector_size: int = 512
@@ -61,7 +61,7 @@ class GPT(BlockEntry):
 
         for part_id, part in node["partitions"].items():
             label = part_id
-            entry_obj, gpt_type, gpt_guid = self._process_entry(part)
+            entry_obj, gpt_type, gpt_guid = self._process_entry(part, sector_size=self._sector_size)
             self._partitions.append(GPTPartition(label, gpt_type, gpt_guid, start=0, size=0, entry=entry_obj))
 
     def size(self) -> int:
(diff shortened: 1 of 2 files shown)


moulin/rouge/ext_utils.py
```diff
@@ -69,9 +69,9 @@ def mkext4fs(file_out: BinaryIO, contents_dir=None):
     _run_cmd(args)
 
 
-def mkvfatfs(file_out: BinaryIO):
-    "Create ext4 fs in given file"
-    args = ["mkfs.vfat", file_out.name]
+def mkvfatfs(file_out: BinaryIO, sector_size: int):
+    "Create FAT fs in given file"
+    args = ["mkfs.vfat", f"-S {sector_size}", file_out.name]
```
lorc, line 74: This should be the optional parameter. We don't want to force default value there.

moulin/rouge/block_entry.py
```diff
@@ -313,6 +315,10 @@ def write(self, fp, offset):
 class Vfat(FileSystem):
     "Represents vfat fs with list of files"
 
+    def __init__(self, node: YamlValue):
+        super(Vfat, self).__init__(node)
+        self._sector_size: int = 0
```
lorc, line 320: `__init__` should initialize the whole object, along with a sector size.

I am proposing to move `sector_size` parameter into `kwargs`

moulin/rouge/block_entry.py
```diff
@@ -62,6 +62,8 @@ def __init__(self, node: YamlValue):
         for part_id, part in node["partitions"].items():
             label = part_id
             entry_obj, gpt_type, gpt_guid = self._process_entry(part)
+            if isinstance(entry_obj, Vfat):
```
lorc, line 65: Okay, this is not a good idea. It is better to pass sector size as a keyword argument to an entry constructor (see below).  So, each `BlockEntry` implementation could make own judgement if and how to use this parameter

moulin/rouge/ext_utils.py
sa-kib: done

moulin/rouge/block_entry.py
sa-kib: yes, moved

moulin/rouge/block_entry.py
sa-kib: fixed

lorc (approved):
Thanks!

PR #81, opened 2023-07-03
rouge: mkvfatfs() hotfix

sa-kib:
f-strings does not work well for mkfs.vfat external ulility. E.g. on Ubuntu 22.04 these failures observed:

[..]
	[INFO] Running mkfs.vfat -S 4096 /tmp/tmpysgqa_7m
	mkfs.fat 4.2 (2021-01-31)
	Bad logical sector size :  4096
	Usage: mkfs.vfat [OPTIONS] TARGET [BLOCKS]
[..]

Neighbour code for similar cases just uses double .append(), so we'd better stick with this approach.

### DIFF
--- a/moulin/rouge/ext_utils.py
+++ b/moulin/rouge/ext_utils.py
@@ -73,7 +73,8 @@ def mkvfatfs(file_out: BinaryIO, sector_size=None):
     "Create FAT fs in given file"
     args = ["mkfs.vfat"]
     if sector_size:
-        args.append(f"-S {sector_size}")
+        args.append("-S")
+        args.append(str(sector_size))
     args.append(file_out.name)
 
     _run_cmd(args)


moulin/rouge/ext_utils.py
```diff
@@ -73,7 +73,8 @@ def mkvfatfs(file_out: BinaryIO, sector_size=None):
     "Create FAT fs in given file"
     args = ["mkfs.vfat"]
     if sector_size:
-        args.append(f"-S {sector_size}")
+        args.append("-S")
+        args.append(str(sector_size))
```
lorc, line 77: Should you just use `args.append(f"-S{sector_size}")` then? 

I can't see how string interpolation alone can cause this issue.

lorc:
@sa-kib could you please rewrite the commit description? As we agreed in Slack, your fix is viable, but you really need to clarify what was the issue, as it is not caused by f-strings per se.

lorc (approved):

PR #82, opened 2023-07-07
rouge: generate target for LZMA packed image (.xz)

rshym:
Archives packed by LZMA (`xz`) can be read without unpacking the full archive.
External tools may use this feature to flash images directly from the *.xz file.

### DIFF
--- a/docs/usage-notes.rst
+++ b/docs/usage-notes.rst
@@ -22,8 +22,8 @@ will also generate handy :code:`image-{image_name}` rules. They can be
 used to invoke :code:`rouge` with the same build options, as
 :code:`moulin` was invoked.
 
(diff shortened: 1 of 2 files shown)


rshym:
Tested with few commands:
```
- ninja -t targets
- ninja full.img.xz
```
Both commands work as expected.

lorc (approved):

rshym:
`Ack` is added. Tested. Ready to be merged.

rshym:
This PR adds functionality that is 'nice to have' for a script with very narrow usage.
So, I'm closing PR, as there is no reason to increase the complexity of the moulin for the feature that will not be used.

rshym:
Thanks, but I decided that this PR is not needed really. The existing function to create .gz is enough for the transfer of the image to the remote host.

PR #83, opened 2023-10-19
builders: Added escape of 'env' parameter

Mishytca:
There is an issue with passing the "EXT_MODULES" parameter from the configuration moulin yaml file to the build.sh context.

The 'android_kernel' builder didn't escape double quotes, while wrapping the 'env' parameter in additional double quotes like this:

bash -c "...env..."

As a result, an incorrect syntax was generated in the build.ninja file.

To generate a functional build.ninja file, it was necessary to take this aspect into account in the moulin.yaml file and add additional escaping there.

This approach is ambiguous since it requires the user to know the implementation details of each moulin builder.

To address this issue, a simple escape function was added to all builders that support the 'env' parameter. It escapes double quotes to ensure proper handling of the 'env' variable.

The issue has been resolved for the following builders:
- android_kernel builder
- android builder
- zephyr builder


svlad-90:
Hi, @lorc, could you, please, have a look? ))

moulin/builders/escape.py
```diff
@@ -0,0 +1,9 @@
+import re
+
+def escape(val):
```
lorc, line 3: It would be nice to have the typing hints and doc string.

moulin/builders/escape.py
```diff
@@ -0,0 +1,9 @@
...
+    result = val
+    result = result.replace(r"\"", "\\\\\\\"")
```
lorc, line 5: Could you use raw string (with `r` prefix) for the second parameter as well? It will reduce number of slashes

moulin/builders/escape.py
Mishytca: Hi @lorc, done.

moulin/builders/escape.py
Mishytca: Hi @lorc, done.

moulin/builders/escape.py
lorc, line 1: `builders` directory is for builders. Python modules from this directory are automatically loaded when `moulin` spots a builder name in YAML file. If someone will write `builder: escape` in YAML, `moulin` will try to include this file, which will lead to spectacular fault. 

The same stands for `fetchers` dir.

lorc (changes requested):

moulin/builders/android_kernel.py
```diff
@@ -8,6 +8,7 @@
 from typing import List
 from moulin.yaml_wrapper import YamlValue
 from moulin import ninja_syntax
+from moulin.builders import escape
```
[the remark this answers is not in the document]
Mishytca: Hi @lorc, done.

moulin/build_conf.py
```diff
@@ -39,7 +39,7 @@ def apply_overrides(self, node: MappingNode):
     @staticmethod
     def _override(remote_node: Node, local_node: Node):
         # pylint: disable=unidiomatic-typecheck
-        if type(remote_node) != type(local_node):
+        if not isinstance(remote_node, type(local_node)):
```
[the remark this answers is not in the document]
Mishytca: Hi @lorc, done.

moulin/build_conf.py
```diff
@@ -39,7 +39,7 @@ def apply_overrides(self, node: MappingNode):
     @staticmethod
     def _override(remote_node: Node, local_node: Node):
         # pylint: disable=unidiomatic-typecheck
```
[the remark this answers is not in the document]
Mishytca: Hi @lorc, done.

moulin/builders/escape.py
Mishytca: Hi @lorc, done.

PR #84, opened 2023-10-25
fetchers: disable prompt usage by git

rshym:
This commit addresses issue with git interacting with the user through the console. For example, HTTPS access to the private repo.
We use git inside the console, that is hidden from the user, so user see only hanging fetch.

So, we instruct the git to not interact with the user by the command `GIT_TERMINAL_PROMPT=0`. As result, git fails cloning and inform user that asking name/password is not possible.
This should push the user to use script-friendly authorization, like ssh.

Suggested-by: Artem Mygaiev <artem_mygaiev@epam.com>

### DIFF
--- a/moulin/fetchers/git.py
+++ b/moulin/fetchers/git.py
@@ -18,14 +18,21 @@ def get_fetcher(conf: YamlValue, build_dir: str, generator: ninja_syntax.Writer)
 
 def gen_build_rules(generator: ninja_syntax.Writer):
     """Generate build rules using Ninja generator"""
-    # If $git_url uses ssh protocol and is not in .ssh/known_hosts,
-    # the ssh client starts to iterate with a user.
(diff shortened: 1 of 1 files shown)


rshym:
Tested with https access to the private repo.

```
FAILED: yocto/test 
GIT_SSH_COMMAND='ssh -o BatchMode=yes' GIT_TERMINAL_PROMPT=0 git clone -q https://github.com/rshym/test.git yocto/test && touch yocto/test
fatal: could not read Username for 'https://github.com': terminal prompts disabled
ninja: build stopped: subcommand failed.
```

rshym:
Rebased on the latest changes.

lorc:
Thanks

rshym:
`Ack` is added.
Tested.
Ready to merge.

PR #87, opened 2023-11-01
yocto.py: Support aliases for specifying 'layers' 

Mishytca:
```
...
      common_layers: &COMMON_LAYERS
            - "layer1"
            - "layer2"
...
components:
    comp0:
         builder:
             layers:
                 - *COMMON_LAYERS
                 - "layer3"
                 - "layer4"
```

The issue is that if an alias appears inside the 'layers' key, the
alias reaches us as an empty node without a value but with a child.
It's the same as writing:

```
layers:
     -
        - "nested_layer1"
        - "nested_layer2"
     - "layer3"
     - "layer4"
```

And there is no way to step over this on the yaml level, e.g. using
some other syntax.

To make the "layers" component combinable with a variable, as in
the example above, the changes were made to the yocto.py file.
The function _flatten_layers has been added. It processes an input YAML
structure, extracting the names of individual layers as strings and
appending them to the resulting list.
The function will return:
['nested_layer1', 'nested_layer2', 'layer3', 'layer4']

Now the above empty node is unfolded for this specific case, and the
proper data is fetched by the moulin.

### DIFF
--- a/moulin/builders/yocto.py
+++ b/moulin/builders/yocto.py
@@ -93,7 +93,7 @@ def _flatten_yocto_conf(conf: YamlValue) -> List[Tuple[str, str]]:
     result: List[Tuple[str, str]] = []
     for entry in conf:
         if not entry.is_list:
-            raise YAMLProcessingError("Exptected array on 'conf' node", entry.mark)
+            raise YAMLProcessingError("Expected array on 'conf' node", entry.mark)
         if entry[0].is_list:
             result.extend([(x[0].as_str, x[1].as_str) for x in entry])
         else:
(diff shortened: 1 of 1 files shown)


moulin/builders/yocto.py
```diff
@@ -154,7 +154,14 @@ def gen_build(self):
         layers_node = self.conf.get("layers", None)
         if layers_node:
             layers_stamp = create_stamp_name(self.yocto_dir, self.work_dir, "yocto", "layers")
-            layers = " ".join([x.as_str for x in layers_node])
+            layers = ""
+            for x in layers_node:
```
lorc, line 158: While this should do the trick, I would like to see this is an separate function.
Also, appending strings in a loop is considered as a bad practice, because string object will be re-created at each iteration, which is bad for performance. I am not saying that we are caring about performance in this particular case, but anyways, it is better do develop good habits :)

Summing up, I'd like to see helper with signature  `_flatten_layers(conf: YamlValue) -> List[str]`

And just `layers = " ".join(_flatten_layers(layers_node))` here

PR #89, opened 2023-11-13
Zephyr builder: Add escaping for the 'vars' parameter

Mishytca:
Fix the issue with passing the "vars" parameters from the configuration moulin YAML file to the build.ninja context.

The 'vars' parameter of the Zephyr builder is placed into the bash -c "... cmd ..." context. It requires an additional level of escaping to be added to the 'vars' parameter.

But that is not enough. The 'vars' parameter is passed to the 'west build' call. West, in turn, passes those 'vars' parameter to CMake.

On the CMake level, Zephyr splits all incoming CMake parameters into 2 parts:

1. Those, which start with the "CONFIG_" prefix
2. All the others

The first group of parameters is stored in the KConfig file. When users pass a string to Kconfig through CMake, they are expected to pass it so that quotes are correct seen from Kconfig, that is:

> cmake -DCONFIG_A_STRING=\"foo bar\"

In this case, to obtain the required level of escaping:

- the first group of parameters should be additionally escaped twice on the moulin level
- the second group of parameters should be additionally escaped once on the moulin level

As a result of this patch, the user can specify the 'vars' on the yaml level in the following way:

        vars:
          - "CONFIG_KERNEL_BIN_NAME=\"%{KERNEL_BIN_NAME}\""
          - "MY_VARIABLE=\"1234567890\""


moulin/builders/zephyr.py
```diff
@@ -75,6 +75,13 @@ def gen_build(self):
         vars_node = self.conf.get("vars", None)
         if vars_node:
             vars_vals = [x.as_str for x in vars_node]
+            for el, var in enumerate(vars_vals):
```
lorc, line 78: I am okay with idea, but implementation is quite unpythonic. I'd prefer to have the next `if` block into a separate helper function. Something akin to `def _escape_env_var(var: str) ->str`. And here, instead of this `for` loop just write `vars_vals = [_escape_env_var(x.as_str) for x in vars_node]`

moulin/builders/zephyr.py
```diff
@@ -75,6 +75,13 @@ def gen_build(self):
...
+                if var.startswith("CONFIG_"):
+                    # If an item in 'vars' starts with "CONFIG_", adding a double quotation mark escaping,
+                    # following the Zephyr documentation
```
lorc, line 81: It would be nice to have reference to that documentation part

moulin/builders/zephyr.py
Mishytca: Hi, @lorc! Done.

moulin/builders/zephyr.py
Mishytca: Hi, @lorc! Done.

docs/conf.py
```diff
@@ -21,7 +21,7 @@
 author = 'EPAM Systems'
 
 # The full version, including alpha/beta/rc tags
-release = 'v0.20'
+release = 'v0.21'
```
lorc, line 24: This should be in a separate commit

moulin/builders/zephyr.py
```diff
@@ -102,6 +102,26 @@ def gen_build(self):
 
         return targets
 
+    def escape_vars_vals(self, var: str) -> str:
```
lorc, line 105: Any particular reason, why this should be a public non-static member of the class? As I can see, it does not touch `self` at all, also I see no reason to make it public.

moulin/builders/zephyr.py
```diff
@@ -102,6 +102,26 @@ def gen_build(self):
...
+
+        Parameters:
+        - var (str): The variable value to be escaped.
+
+        Returns:
+        - str: The escaped variable value.
+
+        If the variable value starts with "CONFIG_", double quotation mark escaping is added,
+        following the:
+        https://github.com/tejlmand/zephyr/commit/4cb2ef83bfd1a4db3c37e66e1cf65bb6ec0145ef
```
lorc, line 117: Can we have link to the official Zephyr repo? I don't know who `tejlmand` are, but there is a chance that they can archive/remove their account and we'll get a dangling link.

lorc (changes requested):
Please remember that commits should be atomic.

docs/conf.py
Mishytca: Hi, @lorc! Done.

moulin/builders/zephyr.py
Mishytca: Hi, @lorc! Done.

moulin/builders/zephyr.py
Mishytca: Hi, @lorc! Done.

Issue #91, opened 2023-11-20
Moulin: Fix the issue when we can't clone two repositories with the same URL but different revisions even when they are clone into different directories


lorc:
This issues was introduced by #86

lorc:
Fixed with #92

PR #92, opened 2023-11-20
git.py: added directory tracking to the Moulin git builder

Mishytca:
Change the behavior of the git fetcher so that it doesn't consider two
or more occurrences of the same git repository in the moulin *.yaml
configuration file as an error if the duplicated repository is cloned
into different sub-folders within the build directory.
For example:
 
```
  domd:
     sources:
      - type: git
        url: "https://github.com/xen-troops/meta-example.git"
        rev: ABCDEF
        dir: "dir1"
  domu:    
    sources:
      - type: git
        url: "https://github.com/xen-troops/meta-example.git"
        rev: branch_A
        dir: "dir2"
```

Above, identical repositories have different branches. But, also, they
have different 'dir' parameters. So that would not be considered as an
error.
 
But if they would have identical directories - the error still will be
triggered.

### DIFF
--- a/moulin/fetchers/git.py
+++ b/moulin/fetchers/git.py
@@ -75,7 +75,8 @@ class GitFetcher:
         if checkout_stamp in _SEEN_REPOS_REV:
             if self.git_rev != _SEEN_REPOS_REV[checkout_stamp]:
                 # Fail on occurrence of different revision for the already downloaded repository
-                raise YAMLProcessingError(f"ERROR: Repository {self.url} has two revisions '{self.git_rev}' "
+                raise YAMLProcessingError(f"ERROR: Repository {self.url} cloned to '{clone_target}' "
+                                          f"has two revisions '{self.git_rev}' "
                                           f"and '{_SEEN_REPOS_REV[checkout_stamp]}'", self.conf["rev"].mark)
             else:
                 # Do not checkout repos for the second time
--- a/moulin/fetchers/git.py
+++ b/moulin/fetchers/git.py
@@ -69,8 +69,8 @@ class GitFetcher:
     def gen_fetch(self):
         """Generate instruction to fetch git repo"""
         clone_target = self.git_dir
-        clone_stamp = create_stamp_name(self.build_dir, self.url, "clone")
-        checkout_stamp = create_stamp_name(self.build_dir, self.url, "checkout")
+        clone_stamp = create_stamp_name(self.build_dir, clone_target, self.url, "clone")
+        checkout_stamp = create_stamp_name(self.build_dir, clone_target, self.url, "checkout")
 
         if checkout_stamp in _SEEN_REPOS_REV:
             if self.git_rev != _SEEN_REPOS_REV[checkout_stamp]:


moulin/fetchers/git.py
```diff
@@ -71,17 +71,19 @@ def gen_fetch(self):
...
-            if self.git_rev != _SEEN_REPOS_REV[checkout_stamp]:
-                # Fail on occurrence of different revision for the already downloaded repository
-                raise YAMLProcessingError(f"ERROR: Repository {self.url} has two revisions '{self.git_rev}' "
-                                          f"and '{_SEEN_REPOS_REV[checkout_stamp]}'", self.conf["rev"].mark)
+        key = (checkout_stamp, clone_target)
+        if key in _SEEN_REPOS_REV:
+            if self.git_rev != _SEEN_REPOS_REV[key]:
+                # Fail on occurrence of different revision for the already downloaded
+                # repository
+                raise YAMLProcessingError(f"ERROR: Repository {self.url} cloned to '{self.git_dir}'"
```
lorc, line 79: You already have `clone_target`, why use `self.git_dir`?

moulin/fetchers/git.py
```diff
@@ -71,17 +71,19 @@ def gen_fetch(self):
         clone_target = self.git_dir
         clone_stamp = create_stamp_name(self.build_dir, self.url, "clone")
         checkout_stamp = create_stamp_name(self.build_dir, self.url, "checkout")
-
-        if checkout_stamp in _SEEN_REPOS_REV:
-            if self.git_rev != _SEEN_REPOS_REV[checkout_stamp]:
-                # Fail on occurrence of different revision for the already downloaded repository
-                raise YAMLProcessingError(f"ERROR: Repository {self.url} has two revisions '{self.git_rev}' "
-                                          f"and '{_SEEN_REPOS_REV[checkout_stamp]}'", self.conf["rev"].mark)
+        key = (checkout_stamp, clone_target)
```
lorc, line 74: Didn't know that `dict` support tuples as keys. Neat trick!

moulin/fetchers/git.py
```diff
@@ -71,17 +71,19 @@ def gen_fetch(self):
...
+        if key in _SEEN_REPOS_REV:
+            if self.git_rev != _SEEN_REPOS_REV[key]:
+                # Fail on occurrence of different revision for the already downloaded
+                # repository
+                raise YAMLProcessingError(f"ERROR: Repository {self.url} cloned to '{self.git_dir}'"
+                                          f"has two revisions '{self.git_rev}' "
+                                          f"and '{_SEEN_REPOS_REV[key]}'", self.conf["rev"].mark)
             else:
                 # Do not checkout repos for the second time
                 return checkout_stamp
```
lorc, line 84: Okay, I genuinely don't understand how it failed before and how your change fixed it. 

Here we are returning `ckeckout_stamp` for Ninja do use as dependency for the next build rule. `checkout_stamp` already depends on `self.build_dir` but not depends on `dirname`, so there are 50% chance that we will return wrong stamp if there are two different repos in the same `build_dir`. 

Could you please explain in details use case where it failed and show how it was fixed?

moulin/fetchers/git.py
Mishytca: Hi @lorc, done

moulin/fetchers/git.py
Mishytca: Hi @lorc, done

moulin/fetchers/git.py
Mishytca: Hi @lorc, done

moulin/fetchers/git.py
```diff
@@ -69,12 +69,13 @@ def __init__(self, conf: YamlValue, build_dir: str, generator: ninja_syntax.Writ
...
-        clone_stamp = create_stamp_name(self.build_dir, self.url, "clone")
-        checkout_stamp = create_stamp_name(self.build_dir, self.url, "checkout")
+        clone_stamp = create_stamp_name(self.build_dir, clone_target, self.url, "clone")
+        checkout_stamp = create_stamp_name(self.build_dir, clone_target, self.url, "checkout")
 
         if checkout_stamp in _SEEN_REPOS_REV:
             if self.git_rev != _SEEN_REPOS_REV[checkout_stamp]:
-                # Fail on occurrence of different revision for the already downloaded repository
+                # Fail on occurrence of different revision for the already downloaded
+                # repository
```
lorc, line 78: Spurious change?

Mishytca: Hi @lorc, done

lorc:
Fixed #91

PR #93, opened 2023-12-04
[WIP] zephyr: add ext_files option to builder

rshym:
This option allows to collect external files that
are required for the build.
Example of corresponding yaml:

```
  builder:
    type: zephyr
    ...
    ext_files:
      "file_1": "path/to/external/file_1"
      "abc/file_BB": "path/to/external/file_2"
      "abc/": "path/to/external/file_3"
```

Destination subdirectories are created if specified. File may be renamed to desired destination name.

### DIFF
--- a/moulin/builders/zephyr.py
+++ b/moulin/builders/zephyr.py
@@ -5,7 +5,7 @@ Zephyr build generator
 """
 
 import os.path
-from typing import List
+from typing import List, cast
 from moulin.yaml_wrapper import YamlValue
 from moulin import ninja_syntax
 from moulin.utils import construct_fetcher_dep_cmd
@@ -28,6 +28,7 @@ def gen_build_rules(generator: ninja_syntax.Writer):
         # Generate fetcher dependency file
         construct_fetcher_dep_cmd(),
         "cd $build_dir",
+        "$pull_ext_sources",
         "source zephyr/zephyr-env.sh",
         "$env west build -p auto -b $board -d $work_dir $target -- $shields $vars",
     ])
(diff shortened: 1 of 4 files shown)


rshym:
Add commit that bumps the version to 0.22, as far as changes made to Zephyr builder require a specific version of moulin.

moulin/builders/zephyr.py
```diff
@@ -56,6 +57,27 @@ def __init__(self, conf: YamlValue, name: str, build_dir: str, src_stamps: List[
...
 
+        list_of_commands = []
+        ext_deps = []
+        ext_files_node = self.conf.get("ext_files", None)
+        if ext_files_node:
+            # if 'ext_files' node is present, then we:
+            # - collect filenames to `ext_deps` to add them to build dependencies
+            # - create destination directories, if required
+            # - copy external files to build-dir (or destination subdirectory, is specified)
+            for dest_name, src_node in cast(YamlValue, ext_files_node).items():
```
lorc, line 68: Do you really need to use `cast` here?

moulin/builders/zephyr.py
```diff
@@ -28,6 +28,7 @@ def gen_build_rules(generator: ninja_syntax.Writer):
         # Generate fetcher dependency file
         construct_fetcher_dep_cmd(),
         "cd $build_dir",
+        "$pull_ext_sources",
```
lorc, line 31: With this approach ninja will copy files every time, even if they are not changed. This is not desired, as main idea of using Ninja (vs plain scripts) is dependency tracking and executing tasks only if they are really should be executed.

If you leave this as is, Zephyr's build system will see that external files timestamps are changed every time, which in turn will trigger rebuild of Zephyr even if it is not needed.  

Long story short, you need to create a new rule and and targets so Ninja will copy files only whey they were updated.

rshym:
This is the second attempt to push this change into the main line. And the second failure.
Considering that it's always better to use the link to some external file than to copy it, we can consider idea, behind this PR, is not a good idea.
So, as a resume - this PR has significant issues with implementation and introduces a bad approach.
I'm closing PR.

PR #94, opened 2023-12-05
yocto.py: layers not being removed from bblayers.conf

Mishytca:
If after building a component (e.g., dom0), layers in the Yaml file are modified, these changes will not be reflected in bblayers.conf. This fix addresses this issue.
Now, if you remove or add layers in the Yaml file, these changes will also be propagated to the bblayers.conf file.

Additionally, the problem with default layers in Poky has been resolved. Poky adds the following layers at its level: 'meta'
"meta-poky"
"meta-yocto-bsp"
More details can be found here:
https://github.com/yoctoproject/poky/blob/
807831067405a465886593df4e3057d3846a0001/documentation/ migration-guides/migration-1.3.rst#bblayersconf

Without this fix, if the above layers are explicitly specified in the Yaml file, an error will occur during the build process.

To address this, the "_filter_yocto_core_layers" function has been added. The function checks the list of layers. If the list contains default Poky layers, the user will receive a warning. Then, the function will removes the default Poky layers from the list, allowing the build to proceed without errors.

Note:
Please, be aware, that the warning in "_filter_yocto_core_layers" function will be changed to an exception soon. Please, adapt your YAML Moulin configuration files.


moulin/builders/yocto.py
```diff
@@ -206,11 +256,16 @@ def gen_build(self):
...
                              "yocto_update_conf",
-                             layers_stamp if layers_node else env_target,
+                             env_target,
                              variables=dict(common_variables, conf=" ".join(local_conf_lines)))
         self.generator.newline()
 
-        self.generator.build(f"conf-{self.name}", "phony", local_conf_target)
+        # Add layers stamp to the dependencies only if it exists
+        dependencies = []
+        dependencies.append(local_conf_target)
```
lorc, line 265: Even C allows declaration and initialization at the same time :)

Instead of this two lines you can just write `dependencies = [local_conf_target]`

Also, to reduce confusion, I propose to rename this variable to `conf_deps` or `conf_dependencies`. Below we already have `deps` variable, so having both `dependencies` and `deps` questions one's sanity :)

moulin/builders/yocto.py
```diff
@@ -221,6 +276,10 @@ def gen_build(self):
         else:
             deps = []
         deps.append(local_conf_target)
+
+        # Add layers stamp to the deps only if it exists
+        if layers_stamp:
```
lorc, line 281: You already have `dependencies`. So what you can do is just `deps.extend(dependencies)`. BTW, I hope that you now see why I proposed to rename `dependencies` :)

lorc (changes requested):
Remember:

1. Changes should be atomic. Warning user about core layers have nothing to do with removing layers from bb-layers file. Split this commit into two.
2. > Describe your changes in imperative mood, e.g. "make xyzzy do frotz"
instead of "[This patch] makes xyzzy do frotz" or "[I] changed xyzzy
to do frotz", as if you are giving orders to the codebase to change
its behaviour

Actually, please read this whole section: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst#n45

3. It is good to make small, localized changes. But please remember that resulting code should be readable and follow internal logic. If you see that your small localized changes are making mess - do not hesitate to rewrite bigger chunks of code.

moulin/builders/yocto.py
```diff
@@ -11,6 +11,11 @@
 from moulin import ninja_syntax
 from moulin.yaml_wrapper import YamlValue
 from moulin.yaml_helpers import YAMLProcessingError
+import logging
+
+
+yocto_core_layers = ["../poky/meta", "../poky/meta-poky", "../poky/meta-yocto-bsp"]
```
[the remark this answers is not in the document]
Mishytca: Hi, @lorc ! Done.

moulin/builders/yocto.py
Mishytca: Hi, @lorc ! Done.

moulin/builders/yocto.py
Mishytca: Hi, @lorc ! Done.

moulin/builders/yocto.py
```diff
@@ -206,12 +220,16 @@ def gen_build(self):
 
         self.generator.build(local_conf_target,
                              "yocto_update_conf",
-                             layers_stamp if layers_node else env_target,
+                             env_target,
                              variables=dict(common_variables, conf=" ".join(local_conf_lines)))
         self.generator.newline()
-
```
lorc, line 212: Stray change

moulin/builders/yocto.py
```diff
@@ -206,12 +220,16 @@ def gen_build(self):
...
         self.generator.build(f"conf-{self.name}", "phony", local_conf_target)
-        self.generator.newline()
```
lorc, line 214: Why are you moving this to  the next `if`?

lorc (changes requested):
Please align the commit message headers with requirements

moulin/builders/yocto.py
```diff
@@ -129,6 +141,43 @@ def _flatten_layers(layers_node: YamlValue) -> List[str]:
     return result
 
 
+def _filter_yocto_core_layers(layers: List[str]) -> List[str]:
```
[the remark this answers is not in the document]
Mishytca: Hi, @lorc ! Done.

moulin/builders/yocto.py
Mishytca: Hi, @lorc ! Done.

Issue #95, opened 2024-01-02
Use dynamic dependecies for the `unpack` fetcher


rshym:
Steps to reproduce
==================

```
wget https://raw.githubusercontent.com/xen-troops/meta-xt-prod-devel-rcar/master/prod-devel-rcar.yaml
moulin ./prod-devel-rcar.yaml --ENABLE_ANDROID yes --ANDROID_PREBUILT_DDK yes
```

Error message appears:
```
moulin.yaml_helpers.YAMLProcessingError: File "rcar-prebuilts-graphics-xt-doma.tar.gz" does not exist   in "./prod-devel-rcar.yaml", line 742, column 23
```

Let's create dummy package with two dummy files:

```
tar -cjf rcar-prebuilts-graphics-xt-doma.tar.gz x1.txt x2.txt
```
And now moulin successfully creates the `build.ninja`.

Reason
======

This archive is required during the creation of the build.ninja to create the following rules:

```
build android/eva/x1.txt android/eva/x2.txt: tar_unpack rcar-prebuilts-graphics-xt-doma.tar.gz
  out_dir = android/eva
build fetch-doma: phony $
    /mnt/t1/builds/yd3test/.stamps/android--https---github.com-xen--troops-android_manifest.git--sync $
    android/eva/x1.txt android/eva/x2.txt
```

Proposed solution
=================

To use dynamic dependencies, as shown in example https://ninja-build.org/manual.html#_tarball_extraction

svlad-90:
It was with @rshym that target issue appears with the following moulin yaml configuration file:
https://github.com/xen-troops/meta-xt-prod-devel-rcar/pull/116/commits/b241827d956c9c7e13a12b2bad27e22c66a3228e

```
  GRAPHICS:
    desc: "Select how to use the GFX (3D hardware accelerator)"
    "binaries":
      overrides:
        variables:
          # for the linux
          # Directory containing ${SOC_NAME}_linux_gsx_binaries_gles.tar.gz
          XT_PREBUILT_GSX_DIR: "${TOPDIR}/../../../prebuilt_gsx"
          # for the android
          XT_DOMA_DDK_KM_PREBUILT_MODULE: "eva/pvr-km/pvrsrvkm.ko"
          XT_DOMA_KERNEL_EXTRA_MODULES: ""
          XT_DOMA_SOURCE_GROUP: "default"
        components:
          domd:
            builder:
              conf:
                - *GFX_CONF
[...]

PR #97, opened 2024-01-17
Dev/mykhailo androsiuk

Mishytca:
Enable building the build.ninja even when the moulin YAML configuration
specifies an archive that is not present in the file system.

Detection of the missing archive error will now occur during the ninja
invocation, specifically when building a ninja target that unpacks the
archive

Added tests are focused on the following issue:
https://github.com/xen-troops/moulin/issues/95

### DIFF
--- a/moulin/fetchers/unpack.py
+++ b/moulin/fetchers/unpack.py
@@ -2,12 +2,20 @@
 # Copyright 2022 EPAM Systems
 """Unpacker fetcher module"""
 
+import sys
 import os.path
 import subprocess
 from typing import List
 from moulin.yaml_helpers import YAMLProcessingError
 from moulin.yaml_wrapper import YamlValue
 from moulin import ninja_syntax
+from moulin.utils import create_stamp_name
+import logging
+
+if __name__ != "__main__":
+    log = logging.getLogger(__name__)
+else:
+    log = logging.getLogger("moulin.fetcher.unpack")
 
 
 def get_fetcher(conf: YamlValue, build_dir: str, generator: ninja_syntax.Writer):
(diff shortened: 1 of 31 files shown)


lorc:
1. Please use minimal YAML files for tests. No need to include whole prod-devel. For example, for my experiments I used this file:
```
desc: "Just a test"

components:
  test:
    builder:
      type: "null"
    sources:
      - type: unpack
        file: test.tar.gz
        archive_type: tar
```

2. I found an issue with your patch: ninja don't want to unpack archive after I touch it. I even added new files to the archive, but ninja refuses to unpack it again. Steps to reproduce:
 - Use the YAML file above
 - Create archive: `tar -czf test.tar.gz ../LICENSE` 
 - Run moulin
 - Run ninja - ninja will unpack archive as expected
 - Re-create archive `tar -czf test.tar.gz ../LICENSE ../pytest.ini` (you can just `touch test.tar.gz` but adding a new file is more revealing)
 - Run ninja - nothing happens, but I expect archive to be re-extracted

3. Before fixing this issue - create a test case that reveals it.

lorc:
Another issue:

```
desc: "Just a test"

components:
  test:
    builder:
      type: "null"
    sources:
      - type: unpack
        file: test.tar.gz
        archive_type: tar
  test2:
    builder:
      type: "null"
    sources:
      - type: unpack
        file: test.tar.gz
        archive_type: tar
```
Result:
```
ninja
[1/1] /home/lorc/work/s4/moulin/moulin.py test.yml
[INFO] Generating build.ninja
ninja: error: build.ninja:47: multiple rules generate .stamps/test.tar.gz.stamp
```

Expected: build succeeds. This is working before your PR.

lorc (commented):
I am terribly sorry, but there is another use case that breaks your patch:

I created the following structure:

```
very_test_dir
very_test_dir/dir with white space
very_test_dir/filename with space
very_test_dir/simple_dir
```
Take note about file names with spaces in them. I saw such occurrences in the wild: Renesas loves to include such files in tar archives with prebuilts. 

Created tar archive of it and...

```
$ ninja
[1/4] Scan test.tar.gz tar archive and create .dd file
ninja: build stopped: multiple rules generate test/with.
```

For your reference:

```
 $ cat .stamps/test-.--test.tar.gz.dd
ninja_dyndep_version = 1
build /home/lorc/work/s4/moulin/tmp/.stamps/test-.--test.tar.gz | test/./very_test_dir/ test/./very_test_dir/dir test/./with test/./white test/./space/ test/./very_test_dir/filename test/./with test/./space test/./very_test_dir/simple_dir/ : dyndep
  restat = 1
```

As always, it would be good to have test case for this.

moulin/fetchers/unpack.py
```diff
@@ -40,33 +52,20 @@ def __init__(self, conf: YamlValue, build_dir: str, generator: ninja_syntax.Writ
         dirname = conf.get("dir", default=".").as_str
         self.out_dir = os.path.join(build_dir, dirname)
         self.type = conf["archive_type"].as_str
-
```
lorc, line 43: Stray change

moulin/fetchers/unpack.py
```diff
@@ -40,33 +52,20 @@ def __init__(self, conf: YamlValue, build_dir: str, generator: ninja_syntax.Writ
...
-        files = self.get_file_list()
-        self.generator.build(files, rule_name, self.fname, variables={"out_dir": self.out_dir})
-
-        return files
+        unpack_rule_name = f"{self.type}_unpack"
+        scan_rule_name = f"{self.type}_scan"
+        stamp = create_stamp_name(self.out_dir, self.fname)
+        dyndep = create_stamp_name(self.out_dir, self.fname) + ".dd"
 
-    def get_file_list(self) -> List[str]:
```
lorc, line 61: Are sure that this does not break `generate_fetcher_dyndep() `function?

BTW, should you use this functionality instead of trying to use bash black in place?

tests/integration_tests/unpack_fetcher/test_tar_archive_duplicate/resources/LICENSE.txt
lorc, line 1: Please rename this to something neutral like "test_file1". No need to confuse future authors.

tests/integration_tests/unpack_fetcher/test_tar_archive_duplicate/resources/pytest.ini
lorc, line 1: Please rename this to something neutral like "test_file2". No need to confuse future authors.

lorc:
Actually, my approach passes all your tests:

```
diff --git a/moulin/fetchers/unpack.py b/moulin/fetchers/unpack.py
index 2393fa6..4216a24 100644
--- a/moulin/fetchers/unpack.py
+++ b/moulin/fetchers/unpack.py
@@ -8,7 +8,7 @@ from typing import List
 from moulin.yaml_helpers import YAMLProcessingError
-
+from moulin.utils import create_stamp_name
 
 def get_fetcher(conf: YamlValue, build_dir: str, generator: ninja_syntax.Writer):
     """Construct and return RepoFetcher object"""
@@ -46,17 +46,13 @@ class UnpackFetcher:
             raise YAMLProcessingError(f"Unkown archive type: {self.type}",
                                       self.conf["archive_type"].mark)
 
-        if not os.path.exists(self.fname):
-            raise YAMLProcessingError(f"File \"{self.fname}\" does not exist",
-                                      self.conf["file"].mark)
-
     def gen_fetch(self) -> List[str]:
         """Generate instructions to unpack archive"""
         rule_name = f"{self.type}_unpack"
-        files = self.get_file_list()
-        self.generator.build(files, rule_name, self.fname, variables={"out_dir": self.out_dir})
+        stamp = create_stamp_name(self.out_dir, self.fname)
+        self.generator.build(stamp, rule_name, self.fname, variables={"out_dir": self.out_dir})
 
-        return files
+        return stamp
 
     def get_file_list(self) -> List[str]:
         "Get list of files in archive"
```

Issue #99, opened 2024-02-18
moulin: add fetcher for custom script builder


al1img:
My custom scripts are located and some repo. I would like to fetch them before execution. `source` node doesn't work for custom build builder.

lorc:
Yeah, this is the problem with custom scripts. AFAIK, current approach is to get the `.yaml` file along with all the custom scripts by cloning a git repo or downloading an archive.

dsemenets:
Created PR https://github.com/xen-troops/moulin/pull/104

PR #101, opened 2024-02-20
builder: custom_script: use src_stamps as dependency

lorc:
Prior to this change "custom_script" builder ignored own source stamp, which made it impossible to use any fetchers for the build. Fix this by adding source stamps to the "deps" list.


lorc:
Should fix #99

PR #102, opened 2024-02-22
Remove the absolute path usage for tools

rshym:
- remove not needed input
- remove abspath from rules

### DIFF
--- a/moulin/build_generator.py
+++ b/moulin/build_generator.py
@@ -104,7 +104,7 @@ def _gen_regenerate(conf_file_name, generator: ninja_syntax.Writer):
     args = " ".join(sys.argv[1:])
     generator.rule("regenerate", command=f"{this_script} {args}", generator=1)
     generator.newline()
-    generator.build(BUILD_FILENAME, "regenerate", [this_script, conf_file_name])
+    generator.build(BUILD_FILENAME, "regenerate", conf_file_name)
     generator.newline()
 
 
--- a/moulin/build_generator.py
+++ b/moulin/build_generator.py
@@ -100,7 +100,7 @@ def generate_fetcher_dyndep(conf: MoulinConfiguration, component: str):
 
 
 def _gen_regenerate(conf_file_name, generator: ninja_syntax.Writer):
-    this_script = os.path.abspath(sys.argv[0])
+    this_script = os.path.basename(sys.argv[0])
     args = " ".join(sys.argv[1:])
     generator.rule("regenerate", command=f"{this_script} {args}", generator=1)
     generator.newline()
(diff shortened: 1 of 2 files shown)


rshym:
This PR may be merged only after https://github.com/xen-troops/meta-xt-prod-devel-rcar/pull/120

Was tested by the build of the devel3.

moulin/build_generator.py
```diff
@@ -100,7 +100,7 @@ def generate_fetcher_dyndep(conf: MoulinConfiguration, component: str):
 
 
 def _gen_regenerate(conf_file_name, generator: ninja_syntax.Writer):
-    this_script = os.path.abspath(sys.argv[0])
+    this_script = os.path.basename(sys.argv[0])
```
lorc, line 103: This helps with development a lot, because it ensures that local moulin is used and not the system one.

moulin/build_generator.py
```diff
@@ -104,7 +104,7 @@ def _gen_regenerate(conf_file_name, generator: ninja_syntax.Writer):
     args = " ".join(sys.argv[1:])
     generator.rule("regenerate", command=f"{this_script} {args}", generator=1)
     generator.newline()
-    generator.build(BUILD_FILENAME, "regenerate", [this_script, conf_file_name])
```
lorc, line 107: It was used for debugging - so Ninja will regenerate build every time Moulin is changed. But now it is obsolete, because Ninja will track only one moulin file, ignoring all modules. So I am okay with the change.

moulin/build_generator.py
rshym: I was thinking about this issue during development, and I agree with you, but partially. Here are my pros and cons for this change, based on which I decided to remove abspath.

Against change:
- abspath allows to use developer's version of the moulin
Pro change:
- we could add the same abspath for rouge as well, for development purposes
- we have a consistent approach for access to all tools inside build.ninja
- during the development we install moulin from the local directory `pip3 install --user -e <development_dir>`

The last option I was using during development of this change (with `pip3 uninstall` just to be sure), and  I found it quite comfortable, taking into account that we can also work with other development things like updates of the version of moulin.

But if you do not agree, I will drop this change. In fact, all this PR is quite a cosmetic change.

moulin/build_generator.py
rshym: Thanks for the clarification, because I was wandering why it's here.

lorc (approved):

rshym:
RB is added.
Tested by `moulin prode-devel-rcar.yaml --MACHINE h3ulcb-4x2g-kf --ENABLE_DOMU yes` with consecutive build.
Ready to be merged.

PR #103, opened 2024-02-26
args argument can be given as value list

dsemenets:


moulin/builders/custom_script.py
```diff
@@ -86,11 +87,16 @@ def gen_build(self):
 
         self.generator.build(f"conf-{self.name}", "phony", local_conf_target)
         self.generator.newline()
-
+        args_node = self.conf.get("args", "")
+        try:
```
lorc, line 91: use of exception for type checking is discouraged. In this case you just can use `is_list` property for `args_node`.

moulin/builders/custom_script.py
```diff
@@ -86,11 +87,16 @@ def gen_build(self):
...
+            args = args_node.as_str
+        except YAMLProcessingError:
+            args_vals = [x.as_str for x in args_node]
+            args = " ".join([val for val in args_vals])
```
lorc, line 95: Why do you do this? `" ".join(args_vals)` will be sufficient. You even can wrap both lines into one:

`args = " ".join([x.as_str for x in args_node])`

moulin/builders/custom_script.py
```diff
@@ -84,7 +84,7 @@ def gen_build(self):
             config_file=local_conf_file))
         self.generator.newline()
 
-        self.generator.build(f"conf-{self.name}", "phony", os.path.join(work_dir, local_conf_target))
+        self.generator.build(f"conf-{self.name}", "phony", local_conf_target)
```
lorc, line 88: Will this break any existing YAML files?

lorc (changes requested):

moulin/builders/custom_script.py
dsemenets: Fixed

moulin/builders/custom_script.py
dsemenets: Fixed

moulin/builders/custom_script.py
dsemenets: No. This file is generated by ninja and passed to script. This is internal deal of build system.

lorc (approved):

PR #106, opened 2024-03-02
custom_script: Make the 'args' parameter optional

dsemenets:
Moulin didn't work if 'args' parameter absent. Make that paremeter optional

### DIFF
--- a/moulin/builders/custom_script.py
+++ b/moulin/builders/custom_script.py
@@ -87,11 +87,14 @@ class CustomScriptBuilder:
         self.generator.build(f"conf-{self.name}", "phony", local_conf_target)
         self.generator.newline()
 
-        args_node = self.conf.get("args", "")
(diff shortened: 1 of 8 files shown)


lorc:
Please fix the spelling in the commit message and add test for all the cases.

lorc:
There is a test framework enabled in Moulin to run automated unit tests:

https://github.com/xen-troops/moulin/tree/main/tests

You need to implement unit tests for `args` handling.

dsemenets:
Added tests

tests/integration_tests/custom_script_builder/script.sh
lorc, line 1: Could you please move this script to the "resources" directory as well?

dsemenets: Done

PR #107, opened 2024-03-04
Add tests for yaml_wrapper

Deedone:
Add tests that ensure that YamlValue and YamlDefaultValue provide the same API, and that the API is working. Fix the bugs that came up during testing.


lorc (commented):
Well done! 

I have one comment: could you please add tests for dict/mapping as well? So we can cover the `YamlWrapper` functionality once and for all.

lorc (approved):

PR #120, opened 2024-10-26
rouge: Add support for resizing raw images

Deedone:
Add new "resize" to the raw_image block, which allows rouge to resize the raw_image to the size set in the block.

This is useful when the image size needs to be changed frequently, or multiple images need to be created from the same source files, as it allows to skip the rootfs rebuild step.

Images are copied to a temporary directory before resizing to preserve the original build artifacts. The temporary directory is created in the current working directory to prevent filling the system's /tmp with large images.

### DIFF
--- a/moulin/rouge/block_entry.py
+++ b/moulin/rouge/block_entry.py
@@ -9,6 +9,7 @@ import struct
 import shutil
 import logging
 import itertools
+import subprocess
 from typing import List, Tuple, NamedTuple, cast
 from tempfile import NamedTemporaryFile, TemporaryDirectory
 
@@ -129,12 +130,14 @@ class RawImage(BlockEntry):
         self._node = node
         self._fname = self._node["image_path"].as_str
         self._size = 0
+        self._resize = True
 
     def _complete_init(self):
         mark = self._node["image_path"].mark
         if not os.path.exists(self._fname):
             raise YAMLProcessingError(f"Can't find file '{self._fname}'", mark)
         fsize = os.path.getsize(self._fname)
+        self._resize = self._node.get("resize", True).as_bool
         size_node = self._node.get("size", None)
         if size_node:
             self._size = _parse_size(size_node)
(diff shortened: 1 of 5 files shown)


Deedone:
@rshym suggested that we should make resize true by default to make behavior more consistent with the other blocks. Will it be a better approach?

rshym:
> @rshym suggested that we should make resize true by default to make behavior more consistent with the other blocks. Will it be a better approach?

After additional discussion, we agreed that 'resize true by default' is not needed.

moulin/rouge/block_entry.py
```diff
@@ -154,7 +156,19 @@ def size(self) -> int:
     def write(self, fp, offset):
         if not self._size:
             self._complete_init()
-        ext_utils.dd(self._fname, fp, offset)
+
+        fsize = os.path.getsize(self._fname)
+
+        if (self._resize and fsize < self._size):
```
lorc, line 162: This is Python, not C :)
 No need for parentheses

lorc (approved):
Good job! With that small nit fixed:

lorc:
> Will it be a better approach?

Yes. I believe in "least surprise" principle. So make it `True` by default.

Deedone:
Also improved the error messaging slightly, I hope it's okay

rshym:
`Reviewed-by: Ruslan Shymkevych <ruslan_shymkevych@epam.com>`

moulin/rouge/ext_utils.py
```diff
@@ -93,3 +93,13 @@ def mmd(img: BinaryIO, folders: list):
     args.extend(folders)
 
     _run_cmd(args)
+
+
```
rshym, line 97: Please remove this empty line

moulin/rouge/block_entry.py
```diff
@@ -154,7 +157,27 @@ def size(self) -> int:
...
+            # Not using default /tmp to prevent filling ram with huge images
+            with TemporaryDirectory(dir=".") as tmpd:
+                shutil.copy(self._fname, tmpd)
+                with open(os.path.join(tmpd, os.path.basename(self._fname)), "rb+") as data:
+                    data.truncate(self._size)
+                try:
+                    ext_utils.resize2fs(os.path.join(tmpd, os.path.basename(self._fname)))
+                except subprocess.CalledProcessError as e:
+                    log.error(
+                        "%s is not resizable. If you don't need to resize it, set 'resize: false' in the yaml.",
```
lorc, line 173: Well, I believe there are multiple reasons why resize2fs may fail.

Thinking of it, this is much more complex task, because raw image could be not only ext2fs, but FAT, for example...

To unblock us, please rewrite the error message like this:

"Failed to resize %s partition. Right now we support resizing for EXT{2,3,4} partitions only. If you don't  really want to resize it, please remove "size" parameter or set "resize" to false. If you want to resize some other type of partitions - please create a PR or notify us at least".

Also, please open issue for mouling about this.

moulin/rouge/ext_utils.py
Deedone: It is required by python styling guides

PR #122, opened 2024-10-29
logs: Add colored output

Deedone:
Make the log output colored depending on the log level. This will make it much easier to quickly spot errors and warnings in the output, and also looks cooler.

Control characters for colors are only added if the output is a TTY, so the output can be redirected to a file without any issues.

Here are some examples:
image
image


lorc (approved):
Great!

PR #127, opened 2025-06-19
Provide a clear message if the default option is not specified

Mishytca:
This change handles the case when "default: true" is not provided
for some option in the YAML file.

Now, in this scenario, a clear "YAMLProcessingError" is raised,
containing detailed and understandable information about the
misconfiguration.

Added pytests for the following scenarios:
-- YAML parameter does not contain a "default: true" variant
(negative case)
-- YAML parameter contains a "default: true" variant (positive case)

Also, the unit test "test_config.py" was updated because
"default: true" is required for all parameters in the YAML
configuration.

PR #129, opened 2025-06-24
Add pytests for validation of missing "default: true" in parameters

Mishytca:
Added pytests for the following scenarios:
-- YAML parameter does not contain a "default: true" variant (negative case)
-- YAML parameter contains a "default: true" variant (positive case)

Also, the unit test "test_config.py" was updated because "default: true" is required for all parameters in the YAML configuration.


Mishytca:
Hi @lorc , please look at this change when you have time.

lorc:
Thanks, but it should be part of #127 . In the same commit, even, as tests should follow changes in the logic. 

Apart from that, tests are looking fine. I'll do a final review when you'll squash this change into #127

Mishytca:
Done. I added this change in https://github.com/xen-troops/moulin/pull/127

PR #130, opened 2025-07-01
Add version check logic comparing local and remote Moulin version

Mishytca:
This change introduces automatic version checking for the Moulin. When Moulin is executed directly (not via ninja), it fetches the remote "setup.py" from GitHub and compares its version with the locally installed one.

If the remote version is newer, a warning is printed with an offer on to upgrade Moulin.

### DIFF
--- a/moulin/main.py
+++ b/moulin/main.py
@@ -16,6 +16,7 @@ import yaml
 from urllib.parse import urlparse, unquote
 import urllib.request
 import os.path
+import json
 
 from packaging.version import Version
 
@@ -136,6 +137,8 @@ def moulin_entry():
         additional_opts=additional_opts)
 
     if not args.fetcherdep:
+        # Check version and notify
+        check_version_and_notify()
         log.info("Generating build.ninja")
         build_generator.generate_build(conf, args.conf)
     else:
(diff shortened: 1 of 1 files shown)


Mishytca:
Hi, @lorc, please take a look at this change when you have time

moulin/main.py
```diff
@@ -248,3 +253,53 @@ def _rouge_validate_output(output_file: str, args):
...
+    # Compare versions and notify
+    if Version(local_version) < Version(remote_version):
+        log.warning("New version %s of Moulin is available. \nLocal version %s. Please update Moulin.",
+                    remote_version,
+                    local_version)
+    else:
+        log.debug("You are using the latest version of Moulin: %s", local_version)
+
+
+def is_invoked_by_ninja() -> bool:
```
lorc, line 295: This function is very fragile. For once, we don't want new external dependencies, like `psutil`.

Second, checking for process name will not guarantee anything. Any process can change its name at any time. 

 Entry points used by ninja are separate, all you need is to call `check_version_and_notify` from user-accessible entry points.

lorc (changes requested):

lorc (approved):
Looks good, thanks

PR #132, opened 2025-08-26
Add utility builders yocto

Mishytca:
Introduce internal utility dispatch and Yocto layer synchronization

This PR adds a generic internal --utility-* dispatch mechanism to Moulin. Utility invocations are detected before normal argument parsing and are forwarded to the corresponding builder-provided handler. The default build configuration flow remains unchanged when no utility option is used.

The Yocto builder now uses this mechanism to synchronize active BitBake layers with the layer list defined in the Moulin YAML manifest.

The Yocto synchronization utility:

- stores the list of Moulin-managed layers in a stamp/state file;
- synchronizes only Moulin-managed layers;
- leaves externally managed layers untouched;
- adds missing YAML layers;
- removes layers that were previously managed but disappeared from the YAML manifest;
- recreates managed layers when needed to preserve the YAML layer order.

Also added unit tests covering:

- Initial synchronization (no stamp file).
- External layers.
- Layer removal.
- Layer addition.
- Layer addition and removal.
- No-op synchronization.
- Layer reordering.
- Layer recreation to preserve YAML order.
- Shell quoting for layer paths containing spaces.

Documentation was updated to describe the internal --utility-* mechanism and clarify that these options are intended for internal use only.

### DIFF
--- a/moulin/builders/yocto.py
+++ b/moulin/builders/yocto.py
@@ -12,6 +12,12 @@ from moulin import ninja_syntax
 from moulin.yaml_wrapper import YamlValue
 from moulin.yaml_helpers import YAMLProcessingError
 import logging
+import sys
+import subprocess
+import argparse
+import os
+import re
+from pathlib import Path
 
 
 YOCTO_CORE_LAYERS = ["../poky/meta", "../poky/meta-poky", "../poky/meta-yocto-bsp"]
@@ -41,16 +47,21 @@ def gen_build_rules(generator: ninja_syntax.Writer):
                    restat=True)
     generator.newline()
 
-    # Add bitbake layers by calling bitbake-layers script
-    cmd = " && ".join([
-        "cd $yocto_dir",
-        "source poky/oe-init-build-env $work_dir",
-        "bitbake-layers add-layer $layers",
-        "touch $out",
-    ])
-    generator.rule("yocto_add_layers",
+    # Minimal CLI: program + --utility-builders-yocto + all args.
+    prog = os.path.basename(sys.argv[0])
+
+    # Add and remove bitbake layers use --utility-builders-yocto
+    cmd = (
+        f"{prog} "
+        "--utility-builders-yocto "
+        "--yocto-dir $yocto_dir "
+        "--work-dir $work_dir "
+        "--layers $layers "
+        "--stamp $out"
+    )
+    generator.rule("yocto_update_layers",
                    command=f'bash -c "{cmd}"',
-                   description="Add yocto layers",
+                   description="Add and remove yocto layers",
                    pool="console",
                    restat=True)
     generator.newline()
@@ -235,7 +246,7 @@ class YoctoBuilder:
             layers_stamp = create_stamp_name(self.yocto_dir, self.work_dir, "yocto", "layers")
             layers = " ".join(_filter_yocto_core_layers(_flatten_layers(layers_node)))
             self.generator.build(layers_stamp,
-                                 "yocto_add_layers",
+                                 "yocto_update_layers",
                                  env_target,
                                  variables=dict(common_variables, layers=layers))
 
@@ -287,3 +298,135 @@ class YoctoBuilder:
         Update stored local conf with actual SRCREVs for VCS-based recipes.
         This should ensure that we can reproduce this exact build later
         """
+
+
+def _env_prefix(yocto_dir: str, work_dir: str) -> str:
+    """
+    Return a shell snippet that cd's into yocto_dir and sources
+    oe-init-build-env for work_dir.
+    """
+    return " && ".join([
+        f"cd {yocto_dir}",
+        f". poky/oe-init-build-env {work_dir}",
+    ])
+
+
+def _run_bash(cmd: str, *, capture=False) -> subprocess.CompletedProcess:
+    """Run a bash -lc command."""
+    return subprocess.run(
+        ["bash", "-lc", cmd],
+        check=True,
+        capture_output=capture,
+        text=capture,
+    )
+
+
+def rels_to_abs(base: Path, rels: List[str]) -> List[str]:
+    return [str((base / rel).resolve(strict=False)) for rel in rels]
+
+
+def handle_utility_call(argv: List[str]) -> int:
+    """
+    Synchronize Yocto build layers with the specification from YAML.
+    The function ensures that the active Yocto layers in a build directory
+    match the layers defined in the YAML configuration. It uses a "stamp"
+    file to detect if synchronization has already been performed.
+
+    Workflow:
+
+      1. If the stamp file does not exist:
+         - Add all specified YAML layers to the build environment.
+         - Create the stamp file.
+
+      2. If the stamp file exists:
+         - Run 'bitbake-layers show-layers' to read the currently active layers.
+         - Normalize their paths and exclude Yocto core layers.
+         - Compare "layers from argv (YAML)" vs "currently active layers."
+         - If the sets differ -> remove extra layers and add missing ones.
+    """
+
+    # Parse args
+    parser = argparse.ArgumentParser(add_help=False)
+    parser.add_argument("--yocto-dir", required=True)  # path to yocto root
+    parser.add_argument("--work-dir", required=True)  # path to build dir
+    parser.add_argument("--layers", nargs="+", required=True)  # YAML layers (relative to yocto root)
+    parser.add_argument("--stamp", required=True)  # stamp file path
+    args = parser.parse_args(argv)
+
+    # Extract args into local variables
+    yaml_layers_str = " ".join(args.layers)
+    yocto_root = args.yocto_dir
+    build_dir = args.work_dir
+    stamp_out = args.stamp
+    stamp_exists = Path(stamp_out).exists()
+    env_prefix = _env_prefix(yocto_root, build_dir)
+
+    # Compute absolute build path
+    build_dir_path = Path(build_dir)
+    build_dir_abs: Path = (
+        build_dir_path if build_dir_path.is_absolute()
+        else (Path(yocto_root) / build_dir_path)).resolve(strict=False)
+
+    # CASE 1: No stamp file yet -> add layers and create stamp
+    if not stamp_exists:
+        _run_bash(" && ".join([
+            f"{env_prefix}",
+            f"bitbake-layers add-layer {yaml_layers_str}",
+            f"touch {stamp_out}",
+        ]))
+        return 0
+
+    # CASE 2: Stamp exists -> check current state and sync if neededs
+    # Run "bitbake-layers show-layers" to list currently active layers
+    # Out of command for examle:
+    #     layer                 path                             priority
+    # ==========================================================================
+    # meta                  /home/../../project/yocto/poky/meta  5
+    # meta-poky             /home/../../project/yocto/poky/meta-poky  5
+    # meta-yocto-bsp        /home/../../project/yocto/poky/meta-yocto-bsp  5
+    # meta-virtualization   /home/../../project/yocto/meta-virtualization  8
+    # meta-oe               /home/../../project/yocto/meta-openembedded/meta-oe  5
+    # meta-filesystems      /home/../../project/yocto/meta-openembedded/meta-filesystems  5
+    # meta-python           /home/../../project/yocto/meta-openembedded/meta-python  5
+    # meta-networking       /home/../../project/yocto/meta-openembedded/meta-networking  5
+    # ...and so on
+    show_output = _run_bash(f"{env_prefix} && bitbake-layers show-layers", capture=True).stdout
+
+    # Extract the 'path' column with regex
+    row_re = re.compile(r"^[ \t]*#?[ \t]*\S+[ \t]+(?P<path>.+?)[ \t]+(?:\d+)[ \t]*$", re.MULTILINE)
+    show_output_paths_only: List[str] = row_re.findall(show_output)
+
+    # Canonicalize build layers path
+    canonical_show_path = [str(Path(p).expanduser().resolve(strict=False)) for p in show_output_paths_only]
+
+    # Get absolute paths for Yocto Core Layers lay
+    yocto_core_layers_abs: List[str] = rels_to_abs(build_dir_abs, YOCTO_CORE_LAYERS)
+
+    # Get absolute paths for YAML-specified layers
+    yaml_layers_abs: List[str] = rels_to_abs(build_dir_abs, args.layers)
+
+    # Filter out core layers to get only build-managed layers
+    used_build_layers_abs = set(canonical_show_path) - set(yocto_core_layers_abs)
+
+    # Determine which layers should be removed
+    to_remove_abs = set(used_build_layers_abs) - set(yaml_layers_abs)
+    # Determine which layers should be added
+    to_add_abs = set(yaml_layers_abs) - set(used_build_layers_abs)
+
+    # If there are layers to remove, otherwise empty string
+    remove_cmd = ("bitbake-layers remove-layer " + " ".join(to_remove_abs)
+                  if to_remove_abs else "")
+
+    # If there are layers to add, otherwise empty string
+    add_cmd = ("bitbake-layers add-layer " + " ".join(to_add_abs)
+               if to_add_abs else "")
+
+    # Always update the stamp file at the end
+    touch_cmd = f"touch {stamp_out}"
+
+    # Collect all command parts, skip empty ones, join with &&
+    cmd_parts = [env_prefix, remove_cmd, add_cmd, touch_cmd]
+    cmd = " && ".join(part for part in cmd_parts if part)
+
+    _run_bash(cmd)
+    return 0
--- a/moulin/builders/yocto.py
+++ b/moulin/builders/yocto.py
@@ -12,6 +12,12 @@ from moulin import ninja_syntax
 from moulin.yaml_wrapper import YamlValue
 from moulin.yaml_helpers import YAMLProcessingError
 import logging
+import sys
+import subprocess
+import argparse
+import os
+import re
+from pathlib import Path
 
 
 YOCTO_CORE_LAYERS = ["../poky/meta", "../poky/meta-poky", "../poky/meta-yocto-bsp", "../openembedded-core/meta"]
@@ -41,16 +47,21 @@ def gen_build_rules(generator: ninja_syntax.Writer):
                    restat=True)
     generator.newline()
 
-    # Add bitbake layers by calling bitbake-layers script
-    cmd = " && ".join([
-        "cd $yocto_dir",
-        "source $distro_dir/oe-init-build-env $work_dir",
-        "bitbake-layers add-layer $layers",
-        "touch $out",
-    ])
-    generator.rule("yocto_add_layers",
+    # Minimal CLI: program + --utility-builders-yocto + all args.
+    prog = os.path.basename(sys.argv[0])
+
+    # Use --utility-builders-yocto to add and remove BitBake layers
+    cmd = (
+        f"{prog} "
+        "--utility-builders-yocto "
+        "--yocto-dir $yocto_dir "
+        "--work-dir $work_dir "
+        "--layers $layers "
+        "--stamp $out"
+    )
+    generator.rule("yocto_update_layers",
                    command=f'bash -c "{cmd}"',
-                   description="Add yocto layers",
+                   description="Synchronize Yocto layers with a moulin configuration file",
                    pool="console",
                    restat=True)
     generator.newline()
@@ -240,9 +251,12 @@ class YoctoBuilder:
         layers_node = self.conf.get("layers", None)
         if layers_node:
             layers_stamp = create_stamp_name(self.yocto_dir, self.work_dir, "yocto", "layers")
-            layers = " ".join(_filter_yocto_core_layers(_flatten_layers(layers_node)))
+            layers = " ".join(
+                shlex.quote(layer)
+                for layer in _filter_yocto_core_layers(_flatten_layers(layers_node))
+            )
             self.generator.build(layers_stamp,
-                                 "yocto_add_layers",
+                                 "yocto_update_layers",
                                  env_target,
                                  variables=dict(common_variables, layers=layers))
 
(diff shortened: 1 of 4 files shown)


moulin/builders/yocto.py
```diff
@@ -41,16 +46,22 @@ def gen_build_rules(generator: ninja_syntax.Writer):
                    restat=True)
     generator.newline()
 
-    # Add bitbake layers by calling bitbake-layers script
+    # Build a stable CLI that reuses the current Moulin invocation (script + args).
+    # This guarantees the utility gets the same build.yaml and global flags.
+    base_cli = shlex.join([os.path.basename(sys.argv[0]), *sys.argv[1:]])
```
lorc, line 51: I don't think that utility invocation needs all arguments, because it does not parse config file. So, `argv[0]` should be sufficient in this case

moulin/builders/yocto.py
```diff
@@ -287,3 +298,113 @@ def capture_state(self):
...
+            # Take the second column (paths)
+            for ln in table_lines[2:]:  # skip the header and "====="
+                s = ln.strip()
+                if not s or set(s) == {"="}:
+                    continue
+                parts = s.split()
+                if len(parts) >= 2:
+                    paths_only.append(parts[1])
+        else:
+            print("Couldn't find table in show-layers output")
```
lorc, line 382: I have a strong feeling that this part can be replaced by a simple regular expression. Also, it would be cool if you add example `bitbake show-layers` output in the comment above, so anyone trying to decipher this code will have some visual clue.

moulin/builders/yocto.py
```diff
@@ -287,3 +298,113 @@ def capture_state(self):
...
+        # 1) Canonicalize (absolute paths, unfold ~, normalize..)
+        canonical_paths = [str(Path(p).expanduser().resolve(strict=False)) for p in paths_only]
+
+        # 2) Compute BUILDDIR absolute path (base for relative layer paths)
+        work_dir_path = Path(args.work_dir)
+        build_abs = (work_dir_path if work_dir_path.is_absolute()
+                     else (Path(args.yocto_dir) / work_dir_path)).resolve(strict=False)
+
+        # 3) Discard the first 3 layers (poky/meta, meta-poky, meta-yocto-bsp)
+        rest = canonical_paths[3:] if len(canonical_paths) > 3 else []
```
lorc, line 393: This is very wild assumption. What if first 3 layers will be different? Let me introduce you to meta-arago: https://layers.openembedded.org/layerindex/branch/master/layer/meta-arago-distro/ which is not based on meta-yocto-bsp as an example.

And why do you need to discard these layers in the first place?

moulin/builders/yocto.py
```diff
@@ -287,3 +298,113 @@ def capture_state(self):
...
+        rel_paths_str = " ".join(p for p in rel_paths)
+
+        # 5) If the number of elements is different, sync_layers immediately
+        if len(args.layers) != len(rel_paths):
+            sync_layers()
+        else:
+            # 6) The quantity is the same --> compare the contents (taking into account the order)
+            desired_rel = [os.path.normpath(p) for p in args.layers]
+            current_rel = [os.path.normpath(p) for p in rel_paths]
+            if desired_rel != current_rel:
```
lorc, line 406: Is there a guarantee that `bitbake show-layers` will print layers in exact same order every time?

moulin/builders/yocto.py
```diff
@@ -287,3 +298,113 @@ def capture_state(self):
...
+    layers_str = " ".join(x for x in args.layers)
+    stamp_out = args.stamp
+    stamp_exists = Path(stamp_out).exists()
+    env_prefix = _env_prefix(args.yocto_dir, args.work_dir)
+
+    def sync_layers():
+        """Remove current layers and add yaml layers; then touch the stamp."""
+        cmd = (
+            f"{env_prefix} && "
+            f"( bitbake-layers remove-layer {rel_paths_str} || true ) && "
```
lorc, line 347: You are referencing to a variable that was not declared prior to this. This is not good.

lorc, line 347: Also, I think you should remove only layers that were removed by user

moulin/builders/yocto.py
```diff
@@ -287,3 +298,113 @@ def capture_state(self):
...
+            f"bitbake-layers add-layer {layers_str} && "
```
lorc, line 348: ... and add layers that were added by user

tests/unittest/test_utility_builders_yocto_conf_missing.py
```diff
@@ -0,0 +1,75 @@
...
+            )
+
+        # check command is not None
+        self.assertIsNotNone(mock_run.call_args)
+
+        # take the command passed to _run_bash
+        show_cmd = mock_run.call_args[0][0]
+
+        # _run_bash called with 'bitbake-layers show-layers'
+        self.assertIn("bitbake-layers show-layers", show_cmd)
```
lorc, line 58: Unit test  should not test for implementation details. It is up to implementation how to get list of layers

tests/unittest/test_utility_builders_yocto_conf_missing.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # cmd contains oe-init-build-env and work_dir
+        self.assertIn(f"cd {yocto_dir}", show_cmd)
+        self.assertIn("oe-init-build-env", show_cmd)
+
+        # sync_layers() -> check the compiled shell command
+        self.assertIsNotNone(mock_subproc.call_args)
+        bash_argv = mock_subproc.call_args[0][0]
+        self.assertEqual(bash_argv[:2], ["bash", "-lc"])
+        sync_cmd = bash_argv[2]
```
lorc, line 68: The same about implementation details

tests/unittest/test_utility_builders_yocto_conf_missing.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # remove empty list + add required layers + touch
+        self.assertIn("bitbake-layers remove-layer  || true", sync_cmd)
+        self.assertIn("bitbake-layers add-layer", sync_cmd)
+        for layer in layers:
+            self.assertIn(layer, sync_cmd)
+        self.assertIn(f"touch {stamp_path}", sync_cmd)
```
lorc, line 75: This is what test should check for: 
1. Unneeded layers were removed
2. Needed layers were added

tests/unittest/test_utility_builders_yocto_first_run.py
```diff
@@ -0,0 +1,62 @@
...
+        # cmd contains bitbake-layers add-layer and var layers
+        self.assertIn("bitbake-layers add-layer", cmd)
+        for layer in layers:
+            self.assertIn(layer, cmd)
+
+        # cmd no contains bitbake-layers show-layers
+        self.assertNotIn("bitbake-layers show-layers", cmd)
+
+        # cmd no contains stamp_path
+        self.assertIn(f"touch {stamp_path}", cmd)
```
lorc, line 62: The same about implementation details. Test should check that correct list of layers was added and stamp was updated.

lorc (changes requested):
You should divide your patches differently:

1. First patch adds generic utility support 
2. Second patch adds yocto's sync layers functionality + tests 

Tests always should be added in the same commit as code that is being tested.

moulin/builders/yocto.py
Mishytca: Hi, @lorc,  I’ve fixed it.

moulin/builders/yocto.py
Mishytca: Hi @lorc, done.

moulin/builders/yocto.py
Mishytca: Hi @lorc.
Thank you for pointing this out.

You're right. The previous implementation relied on a Poky-specific assumption by treating a predefined set of layers as "core" layers. I have removed this assumption from the synchronization utility.

The utility now synchronizes only the layers managed by Moulin. It stores the list of managed layers in the stamp/state file and uses that information during subsequent synchronizations. As a result, externally managed layers remain untouched, regardless of the underlying Yocto distribution (Poky, meta-arago, or any other).

The existing Poky-specific filtering remains in the Yocto builder to preserve the previously implemented behavior for explicit Poky default layers, but the synchronization logic itself no longer depends on any Poky-specific layer classification.

moulin/builders/yocto.py
Mishytca: Hi @lorc, done.

moulin/builders/yocto.py
Mishytca: Hi @lorc, done.

moulin/builders/yocto.py
Mishytca: Hi @lorc, done.

moulin/builders/yocto.py
Mishytca: Hi @lorc, done.

tests/unittest/test_utility_builders_yocto_conf_missing.py
Mishytca: Hi @lorc, done.

tests/unittest/test_utility_builders_yocto_conf_missing.py
Mishytca: Hi @lorc, done.

tests/unittest/test_utility_builders_yocto_first_run.py
Mishytca: Hi @lorc, done.

tests/unittest/test_utility_builders_yocto_conf_missing.py
Mishytca: Hi @lorc, done.

moulin/builders/yocto.py
```diff
@@ -41,16 +47,23 @@ def gen_build_rules(generator: ninja_syntax.Writer):
                    restat=True)
     generator.newline()
 
-    # Add bitbake layers by calling bitbake-layers script
+    # Minimal CLI: program + build.yaml (utility doesn't need global flags).
+    prog = os.path.basename(sys.argv[0])
+    conf_path = sys.argv[1]  # предполагаем, что это build.yaml
```
lorc: Could you please review your patch for yourself very carefully before submitting them?

moulin/builders/yocto.py
```diff
@@ -41,16 +47,23 @@ def gen_build_rules(generator: ninja_syntax.Writer):
                    restat=True)
     generator.newline()
 
-    # Add bitbake layers by calling bitbake-layers script
+    # Minimal CLI: program + build.yaml (utility doesn't need global flags).
+    prog = os.path.basename(sys.argv[0])
+    conf_path = sys.argv[1]
```
lorc: You can't assume that path to config file will be in the first argument. See moulin usage:

`usage: moulin [-h] [--help-config] [-v] [--dump] build.yaml`

Why do you need it anyways?

moulin/builders/yocto.py
```diff
@@ -41,16 +47,23 @@ def gen_build_rules(generator: ninja_syntax.Writer):
...
-        "cd $yocto_dir",
-        "source poky/oe-init-build-env $work_dir",
-        "bitbake-layers add-layer $layers",
-        "touch $out",
+        f"{base_cli} "
+        "--utility-builders-yocto "
+        "--yocto-dir $yocto_dir "
+        "--work-dir $work_dir "
+        "--layers $layers "
+        "--stamp $out",
```
lorc: I am not fully understand why you need `"  && ".join([])` part if you are basically providing single command. Care to elaborate?

moulin/builders/yocto.py
```diff
@@ -41,16 +47,23 @@ def gen_build_rules(generator: ninja_syntax.Writer):
...
     ])
     generator.rule("yocto_add_layers",
                    command=f'bash -c "{cmd}"',
-                   description="Add yocto layers",
+                   description="Add and remove yocto layers",
```
lorc: Probably you need to rename rule to "yocto_update_layers" as well.

moulin/builders/yocto.py
```diff
@@ -287,3 +300,141 @@ def capture_state(self):
...
+
+def rels_to_abs(base: Path, rels: List[str]) -> List[str]:
+    return [str((base / rel).resolve(strict=False)) for rel in rels]
+
+
+def handle_utility_call(conf, argv: List[str]) -> int:
+    """
+        Synchronize Yocto build layers with the specification from YAML.
+        The function ensures that the active Yocto layers in a build directory
+    match the layers defined in the YAML configuration. It uses a "stamp"
```
lorc: Strange formatting of the doc string

moulin/builders/yocto.py
```diff
@@ -287,3 +300,141 @@ def capture_state(self):
...
+    file to detect if synchronization has already been performed.
+
+    Workflow:
+
+      1. If the stamp file does not exist:
+         - Add all specified YAML layers to the build environment.
+         - Create the stamp file.
+
+      2. If the stamp file exists:
+         - Run 'bitbake-layers show-layers' to read the
```
lorc: to read the what?

moulin/builders/yocto.py
```diff
@@ -287,3 +300,141 @@ def capture_state(self):
...
+    # meta                  /home/../../project/yocto/poky/meta  5
+    # meta-poky             /home/../../project/yocto/poky/meta-poky  5
+    # meta-yocto-bsp        /home/../../project/yocto/poky/meta-yocto-bsp  5
+    # meta-virtualization   /home/../../project/yocto/meta-virtualization  8
+    # meta-oe               /home/../../project/yocto/meta-openembedded/meta-oe  5
+    # meta-filesystems      /home/../../project/yocto/meta-openembedded/meta-filesystems  5
+    # meta-python           /home/../../project/yocto/meta-openembedded/meta-python  5
+    # meta-networking       /home/../../project/yocto/meta-openembedded/meta-networking  5
+    # ...and so on
+    show_output = _run_bash(f"{env_prefix} && bitbake-layers show-layers", capture=True).stdout
```
lorc: I don't think this is a good variable name. You are not showing output to anyone

moulin/builders/yocto.py
```diff
@@ -287,3 +300,141 @@ def capture_state(self):
...
+
+    # Extract the 'path' column with regex
+    row_re = re.compile(r"^[ \t]*#?[ \t]*\S+[ \t]+(?P<path>.+?)[ \t]+(?:\d+)[ \t]*$", re.MULTILINE)
+    paths_only: List[str] = row_re.findall(show_output)
+
+    # Canonicalize build layers path
+    canonical_show_path = [str(Path(p).expanduser().resolve(strict=False)) for p in paths_only]
```
lorc: Again, why "show"?

moulin/builders/yocto.py
```diff
@@ -287,3 +300,141 @@ def capture_state(self):
...
+
+    # Parse CLI arguments
+    parser = argparse.ArgumentParser(add_help=False)
+    parser.add_argument("--yocto-dir", required=True)
+    parser.add_argument("--work-dir", required=True)
+    parser.add_argument("--layers", nargs="+", required=True)
+    parser.add_argument("--stamp", required=True)
+    args = parser.parse_args(argv)
+
+    # Extract arguments into local variables
```
lorc: I dunno why you need a bunch of local variables, some of them you use only once, but it is up to you...

moulin/builders/yocto.py
```diff
@@ -287,3 +300,141 @@ def capture_state(self):
...
+    stamp_exists = Path(stamp_out).exists()
+    env_prefix = _env_prefix(yocto_dir, work_dir)
+
+    # Compute absolute build path
+    work_dir_path = Path(work_dir)
+    build_abs: Path = (
+        work_dir_path if work_dir_path.is_absolute()
+        else (Path(yocto_dir) / work_dir_path)).resolve(strict=False)
+
+    # CASE 1: No stamp file yet -> add layers and create stamp
```
lorc: Do you really need this special case? What is the difference with the next one?

moulin/builders/yocto.py
```diff
@@ -287,3 +300,141 @@ def capture_state(self):
...
+
+    spec_layers_abs = yaml_layers_abs
+    actual_layers_abs = used_build_layers_abs
+
+    # Convert yaml layers and build layer lists into set
+    spec_set = set(spec_layers_abs)
+    actual_set = set(actual_layers_abs)
+
+    # Determine which layers should be removed
+    to_remove_abs = [p for p in actual_layers_abs if p not in spec_set]
```
lorc: `to_remove_abs = actual_set - spec_set`, no?

moulin/builders/yocto.py
```diff
@@ -287,3 +300,141 @@ def capture_state(self):
...
+    # Determine which layers should be added
+    to_add_abs = [p for p in spec_layers_abs if p not in actual_set]
```
lorc: Simillary,

`to_add_abs = spec_layers - actual_set`

moulin/builders/yocto.py
```diff
@@ -287,3 +300,141 @@ def capture_state(self):
...
+    # Get absolute paths for Yocto Core Layers lay
+    yocto_core_layers_abs: List[str] = rels_to_abs(build_abs, YOCTO_CORE_LAYERS)
+
+    # Get absolute paths for YAML-specified layers
+    yaml_layers_abs: List[str] = rels_to_abs(build_abs, layers)
+
+    # Filter out core layers to get only build-managed layers
+    used_build_layers_abs = [p for p in canonical_show_path if p not in yocto_core_layers_abs]
+
+    spec_layers_abs = yaml_layers_abs
```
lorc: Why you are introducing even more local variables? This function is already hard to follow...

moulin/builders/yocto.py
```diff
@@ -287,3 +300,141 @@ def capture_state(self):
...
+    spec_set = set(spec_layers_abs)
+    actual_set = set(actual_layers_abs)
+
+    # Determine which layers should be removed
+    to_remove_abs = [p for p in actual_layers_abs if p not in spec_set]
+    # Determine which layers should be added
+    to_add_abs = [p for p in spec_layers_abs if p not in actual_set]
+
+    # If there are layers to remove, otherwise empty string
+    remove_cmd = ("bitbake-layers remove-layer " + " ".join(p for p in to_remove_abs)
```
lorc: isn't `p for p` looks a bit tautological for you? Why do you need loop in here?

moulin/builders/yocto.py
```diff
@@ -287,3 +300,141 @@ def capture_state(self):
...
+                  if to_remove_abs else "")
+    # If there are layers to add, otherwise empty string
+    add_cmd = ("bitbake-layers add-layer " + " ".join(p for p in to_add_abs)
```
lorc: The same tautological thing. 

Probably you wanted to write just `join(to_add_abs)` or `join(list(to_add_abs))` at worst.

lorc (changes requested):
Please revisit use and naming of local variables. `handle_utility_call` is a total mess.

moulin/builders/yocto.py
Mishytca: Hi @lorc, thank you for taking the time. I will recheck everything.

moulin/builders/yocto.py
Mishytca: Hi, @lorc. You are absolutely right. I decided not to use the configuration in the utility.

moulin/builders/yocto.py
Mishytca: Hi @lorc , done.

moulin/builders/yocto.py
Mishytca: Hi @lorc, done.

moulin/builders/yocto.py
Mishytca: Hi, @lorc, done.

moulin/builders/yocto.py
Mishytca: Hi, @lorc. `to_add_abs` can indeed be viewed as spec - actual, but I intentionally keep it as a list comprehension to preserve the YAML order.

moulin/builders/yocto.py
Mishytca: Hi @lorc . ` to_remove_abs` cannot be computed as actual - spec, because actual also contains layers that are not managed by Moulin. We only remove layers that were previously managed by Moulin and are no longer present in the YAML manifest. External layers must remain untouched.

moulin/builders/yocto.py
Mishytca: Hi @lorc, done

moulin/builders/yocto.py
Mishytca: Hi @lorc, done

moulin/builders/yocto.py
Mishytca: Hi @lorc, done

moulin/builders/yocto.py
Mishytca: Hi @lorc. 
Using " && ".join([...]) is not allowed; the invocation must be a single command. Thanks for noticing this.

Mishytca:
> Please revisit use and naming of local variables. `handle_utility_call` is a total mess.
I reviewed the local variable names, tried to make the function simpler and more readable, and added several comments to make it clearer.

klogg:
@Mishytca can you update please?

moulin/builders/yocto.py
```diff
@@ -287,3 +298,135 @@ def capture_state(self):
...
+               if to_add_abs else "")
+
+    # Always update the stamp file at the end
+    touch_cmd = f"touch {stamp_out}"
+
+    # Collect all command parts, skip empty ones, join with &&
+    cmd_parts = [env_prefix, remove_cmd, add_cmd, touch_cmd]
+    cmd = " && ".join(part for part in cmd_parts if part)
+
+    _run_bash(cmd)
```
svlad-90, line 520: Hi @Mishytca, will the "bitbake-layers show-layers" output after this command always match (including order) the layers listed in the moulin manifest YAML?

tests/unittest/test_utility_builders_yocto_sync_layers.py
```diff
@@ -0,0 +1,124 @@
+import unittest
```
svlad-90, line 1: Hi @Mishytca,

The PR description lists four scenarios, but the tests cover only two: the first run without a stamp, and a run with a stamp that removes/adds. There is no separate test for “stamp exists, conf missing / only core layers”, no test for “identical layers -> no add/remove”, and no test for ordering.

moulin/builders/yocto.py
```diff
@@ -287,3 +298,135 @@ def capture_state(self):
         Update stored local conf with actual SRCREVs for VCS-based recipes.
         This should ensure that we can reproduce this exact build later
         """
+
+
+def _env_prefix(yocto_dir: str, work_dir: str) -> str:
```
svlad-90, line 313: Hi @Mishytca,

Affected places: _env_prefix(), first-run bitbake-layers add-layer, remove_cmd, add_cmd, and touch_cmd.

The problem is that paths are inserted directly into the shell string. If yocto_dir, work_dir, a layer path, or the stamp path contains spaces or shell-sensitive characters, the command may break or execute something unintended.

svlad-90 (changes requested):

moulin/main.py
```diff
@@ -30,6 +31,28 @@
 
 OptionDef = Tuple[List[str], Dict[str, Any]]
 
+# Marker prefix for utility calls
+UTILITY_PREFIX = "--utility-"
+
+
+def _dispatch_utility(mod_qual: str, argv: List[str]) -> int:
```
svlad-90, line 38: Hi @Mishytca, please update the documentation regarding the utility mechanism.

Add it to this section:
https://moulin.readthedocs.io/en/latest/user-reference.html#internal-command-line-options

svlad-90 (changes requested):

moulin/builders/yocto.py
Mishytca: Hello, @svlad-90! Thank you for your review and your comments.

I updated the logic to handle this case. Now Moulin checks whether a simple remove/add operation would preserve the YAML order. If not, it recreates the managed layers in the exact order defined by the YAML manifest.

moulin/builders/yocto.py
Mishytca: This has been addressed by quoting all paths before inserting them into the shell command. yocto_dir, work_dir, layer paths, and the stamp path are now wrapped with shlex.quote(), so paths containing spaces or other shell-sensitive characters are handled correctly.

moulin/main.py
Mishytca: The documentation has been updated.

tests/unittest/test_utility_builders_yocto_sync_layers.py
Mishytca: Thank you for pointing this out.

You're absolutely right. At the time of your review, those scenarios were indeed missing.

I've now extended the test coverage to include:

stamp exists with only Yocto core layers present;
stamp exists with identical layers (no add/remove expected);
layer reordering when the managed layer set matches but the order differs;
only add;
only remove;
content differs and requires layer recreation to preserve the YAML order;
shell quoting for layer paths containing spaces.
The PR description has also been updated.

svlad-90:
Reviewed-by: Vladyslav Goncharuk <vladyslav_goncharuk@epam.com>

Mishytca:
Hello, @lorc . I’ve updated the PR. Please review the changes when you have time.

lorc (approved):

Mishytca:
`R-b` are added.
Ready to be merged.
Thank you!

PR #135, opened 2025-09-28
rouge: use pigz when possible

rshym:
The usage of pigz significantly reduces the compression time of the big images. So let's use pigz if available.

### DIFF
--- a/moulin/rouge/__init__.py
+++ b/moulin/rouge/__init__.py
@@ -36,7 +36,8 @@ def gen_build_rules(generator: ninja_syntax.Writer):
     moulin_args = " ".join(sys.argv[1:])
     generator.rule("rouge", f"rouge {moulin_args} -fi $image_name -o $out")
     generator.newline()
-    generator.rule("gzip", "gzip -1kf $in")
+    # the pigz is faster, but gzip is available everywhere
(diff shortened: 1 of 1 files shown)


moulin/rouge/__init__.py
```diff
@@ -36,7 +36,7 @@ def gen_build_rules(generator: ninja_syntax.Writer):
     moulin_args = " ".join(sys.argv[1:])
     generator.rule("rouge", f"rouge {moulin_args} -fi $image_name -o $out")
     generator.newline()
-    generator.rule("gzip", "gzip -1kf $in")
+    generator.rule("gzip", "command -v pigz &> /dev/null && pigz -1kf $in || gzip -1kf $in")
```
lorc: Well, after reading `man command` I got how it is working, but I'd like to see comment like "Try pigz if available or fall back to gzip"

rshym: Comment is added.

After the additional testing, I have found that `ninja` is using `/bin/sh -c` to run commands. And inside our docker `/bin/sh` is link to the `dash`. And `dash` doesn't understand `&>` from the`bash`. Even more, this interferes with the exit code of the `command -v pigz`.
So, I have replaced bashism `&> /dev/null`, with the proper POSIX variant `> /dev/null 2>&1`.

lorc (approved):

rshym:
`Ack` is added.
Tested.
Ready to be merged.

PR #144, opened 2026-06-26
west: pass git options to west init

svlad-90:
Add YAML support for git_options in the Moulin west fetcher. The options are passed to west init as repeated -o=<option> arguments, matching upstream west commit
zephyrproject-rtos/west@4a2c6bba689c656d46cae80889a686b23e9980c1

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -23,7 +23,7 @@ jobs:
     - name: Install dependencies
       run: |
         python3 -m pip install --upgrade pip
-        pip3 install flake8 pytest pyaml
+        pip3 install flake8 pytest pyaml west
         # Needed by pygit package
         sudo apt-get install python3-pygit2
         # Needed by moulin integration tests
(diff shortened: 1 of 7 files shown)


svlad-90:
Verification report:
[west_fetcher_git_options_report.html](https://github.com/user-attachments/files/29387479/west_fetcher_git_options_report.html)

svlad-90:
Regarding the developed test scenario:

This integration test creates local fake Git remotes for both the project repository and the west manifest repository. The project repository has two commits so that shallow cloning can be meaningfully verified.

The test then generates a moulin input YAML pointing to the local manifest remote, runs moulin, and checks that the generated build.ninja contains the expected git option for depth=1.

After that, the test runs the generated ninja target that initializes the west workspace. Finally, it verifies the actual Git state of the cloned manifest repository and checks that it is a shallow repository.

So the test validates the full path from moulin YAML configuration to generated ninja rules and then to the real Git clone result, not only the generated text.

andr2000:

lorc:

PR #145, opened 2026-06-30
Add west group_filter support

svlad-90:
Zephyr west manifests may use project groups to exclude optional
repositories from a workspace. Moulin did not expose west's
manifest.group-filter setting, so every project listed in the manifest was
fetched even when a build only needed a subset of groups.

Add a west fetcher group_filter option. The option accepts either the west
group filter expression as a string or a YAML list of expressions; list
items are joined with commas before being written to west config.

Before running west update, configure the workspace-local
manifest.group-filter value. If group_filter is not specified, remove the
local value to avoid reusing a stale filter from a previous run.

This follows west's manifest group filter semantics:
https://docs.zephyrproject.org/latest/develop/west/manifest.html#project-groups

### DIFF
--- a/docs/user-reference.rst
+++ b/docs/user-reference.rst
@@ -373,6 +373,7 @@ Complete list of supported options:
   file: manifest-file.yml
   git_options:
     - "--depth=1"
+  group_filter: "-group1,+group2"
 
 * :code:`type` - mandatory - should be :code:`west` to enable `west` fetcher.
 * :code:`url` - optional - manifest repository URL. You can provide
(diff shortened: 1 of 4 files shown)


svlad-90:
### Test Scenario Description

  The added integration test for the `west` fetcher to verify `group_filter` support in both supported YAML forms: scalar string and list.

  The test creates a temporary west manifest with two local git projects:

  - `kept`, assigned to group `kept`
  - `dropped`, assigned to group `dropped`

  It runs the same fetch flow for both configurations:

  ```yaml
  group_filter: "+kept,-dropped"

  group_filter:
    - "+kept"
    - "-dropped"
 ```

  Both forms are expected to normalize to:

```
  +kept,-dropped
```

  The test then:

  1. Generates build.ninja through moulin.py.
  2. Verifies that build.ninja contains group_filter = +kept,-dropped.
  3. Runs ninja fetch-test, which executes the generated west fetch pipeline.
  4. Verifies that west config manifest.group-filter is set to +kept,-dropped.
  5. Checks west list output: kept is present and dropped is absent.
  6. Checks the workspace filesystem: workspace/kept/.git exists and workspace/dropped was not created.

lorc:

PR #146, opened 2026-07-02
Add builder dependency tracking

svlad-90:
Fetcher-reported dependency files can be too broad for large source workspaces: a change anywhere in a fetched repository may rebuild components even when the builder did not use that file.

Introduce component dependency policies and the internal `--dep <component>` mode. Keep `--fetcherdep` as the legacy fetcher-only path, add fetched_files/build_files/all_files policies, and generate policy-selected dependency files after successful builds.

Add Zephyr builder support for build-file reporting from CMake/Ninja metadata, update Yocto/custom_script/Zephyr rules to use the new dependency command, document the depfile lifecycle, and cover the behavior with unit and integration tests.


svlad-90:
[builderdep-final-validation-report.html](https://github.com/user-attachments/files/29643958/builderdep-final-validation-report.html)

Local testing results.

moulin/build_generator.py
```diff
@@ -78,25 +78,51 @@ def generate_build(conf: MoulinConfiguration,
...
     builder_type = builder_node["type"].as_str
     builder_module = builder_modules[builder_type]
-    builder = builder_module.get_builder(builder_node, component, build_dir, [], generator)
+    # Dependency-only mode does not generate Ninja rules. Builders that provide
+    # dependency metadata must keep that path independent from rule generation.
+    builder = builder_module.get_builder(builder_node, component, build_dir, [], None)
 
-    deps: List[str] = []
     targets = builder.get_targets()
+    return {
```
lorc: Please do not use dicts to return multiple values. This will cause typing problems. There is `dataclass` for your disposal.

moulin/build_generator.py
```diff
@@ -14,9 +14,18 @@
 from moulin import make_syntax
 from moulin import rouge
 from moulin import yaml_helpers as yh
+from moulin.yaml_helpers import YAMLProcessingError
 from moulin.build_conf import MoulinConfiguration
 
 BUILD_FILENAME = 'build.ninja'
+DEPENDENCY_POLICY_FETCHED = "fetched_files"
```
lorc: Ugh... if you want an enum, there is `enum` module

moulin/build_generator.py
```diff
@@ -84,6 +98,25 @@ def generate_fetcher_dyndep(conf: MoulinConfiguration, component: str):
     _write_dyndep(component, deps_context["targets"], deps)
 
 
+def generate_component_dyndep(conf: MoulinConfiguration, component: str):
+    _flatten_sources(conf)
+
+    deps_context = _get_dependency_context(conf, component)
+    policy = _get_dependency_policy(deps_context["component_node"])
+
+    if policy in [DEPENDENCY_POLICY_BUILD, DEPENDENCY_POLICY_ALL]:
```
lorc: Use tuple instead of list for `in`. There is no need to create mutable object

moulin/build_generator.py
```diff
@@ -115,10 +149,88 @@ def _get_fetcher_file_list(deps_context) -> List[str]:
...
+                                       builder_modules,
+                                       fetcher_modules) -> None:
+    root = conf.get_root()
+    for component_name, component_node in root["components"].items():
+        build_dir = component_node.get("build-dir", component_name).as_str
+        builder_node = component_node["builder"]
+        builder_type = builder_node["type"].as_str
+        builder_module = builder_modules[builder_type]
+        # Dependency validation only needs metadata. Builders that expose
+        # get_build_file_list must make that method independent from generator.
```
lorc: Why do you have duplicate comments? I see exactly the same comment in `_ensure_fetcher_file_support_for_component`

moulin/builders/custom_script.py
```diff
@@ -39,9 +37,11 @@ def gen_build_rules(generator: ninja_syntax.Writer):
                    description="Updating config")
     generator.newline()
 
+    # Keep the build command in a subshell so the following --dep call runs
+    # from Ninja's original working directory and writes .moulin_$name.d there.
```
lorc: I'd prefer `push` and `popd`, honestly

moulin/builders/yocto.py
```diff
@@ -72,12 +72,11 @@ def gen_build_rules(generator: ninja_syntax.Writer):
...
     # Invoke bitbake. This rule uses "console" pool so we can see the bitbake output.
+    # Keep the build command in a subshell so the following --dep call runs
+    # from Ninja's original working directory and writes .moulin_$name.d there.
     cmd = " && ".join([
-        # Generate fetcher dependency file
-        construct_fetcher_dep_cmd(),
-        "cd $yocto_dir",
-        "source $distro_dir/oe-init-build-env $work_dir",
-        "bitbake $target",
+        "( cd $yocto_dir && source $distro_dir/oe-init-build-env $work_dir && bitbake $target )",
```
lorc: `pushd/popd`

moulin/builders/zephyr.py
```diff
@@ -25,12 +25,12 @@ def gen_build_rules(generator: ninja_syntax.Writer):
...
+    # Keep the build command in a subshell so the following --dep call runs
+    # from Ninja's original working directory and writes .moulin_$name.d there.
     cmd = " && ".join([
-        # Generate fetcher dependency file
-        construct_fetcher_dep_cmd(),
-        "cd $build_dir",
-        "source zephyr/zephyr-env.sh",
-        "$env west build -p auto -b $board -d $work_dir $target $snippets -- $shields $vars",
+        "( cd $build_dir && source zephyr/zephyr-env.sh && "
+        "$env west build -p auto -b $board -d $work_dir $target $snippets -- $shields $vars )",
```
lorc: `popd/pushd`

moulin/utils.py
```diff
@@ -14,11 +14,11 @@ def create_stamp_name(*args):
...
 
 
-def construct_fetcher_dep_cmd() -> str:
-    "Generate command line to generate fetcher dependency file"
+def construct_dep_cmd() -> str:
+    "Generate command line to generate component dependency file"
     this_script = os.path.basename(sys.argv[0])
     args = " ".join(sys.argv[1:])
-    return f"{this_script} {args} --fetcherdep $name"
+    return f"{this_script} {args} --dep $name"
```
lorc, line 21: I am not sure I got your intent. In the previous patch you left support for both `--fetcherdep` and `--dep`, but now you are removing function that generates `--fetcherdep` calls. What is the reason to leave other code? It is basically dead now

moulin/builders/zephyr.py
```diff
@@ -136,6 +141,129 @@ def get_targets(self):
         "Return list of targets that are generated by this build"
         return [os.path.join(self.build_dir, t.as_str) for t in self.conf["target_images"]]
 
+    def get_build_file_list(self) -> List[str]:
+        """Return files observed by Zephyr/CMake build metadata.
+
+        Dependency-only code may instantiate builders with generator=None.
```
lorc: I am not sure I got sense of this

moulin/builders/zephyr.py
```diff
@@ -136,6 +141,129 @@ def get_targets(self):
...
+
+    def __get_ninja_deps_files(self, zephyr_build_dir: str) -> List[str]:
+        if not os.path.isdir(zephyr_build_dir):
+            return []
+
+        result = subprocess.run(["ninja", "-C", zephyr_build_dir, "-t", "deps"],
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.DEVNULL,
+                                text=True)
+        if result.returncode != 0:
```
lorc, line 195: It would be nice to print a warning at least. Maybe together with stderr/stdout

moulin/builders/zephyr.py
```diff
@@ -136,6 +141,129 @@ def get_targets(self):
...
+        return paths
+
+    def __get_west_manifest_file(self) -> List[str]:
+        try:
+            result = subprocess.run(["west", "manifest", "--path"],
+                                    cwd=self.build_dir,
+                                    stdout=subprocess.PIPE,
+                                    stderr=subprocess.DEVNULL,
+                                    text=True)
+        except OSError:
```
lorc: Why are you expecting OSError here, but for similar call to Ninja in a function above?

Also, it would be really nice to warn user about problems

lorc (changes requested):
You commit message "Use the Ninja inputs tool when available and query as a Ninja 1.10 fallback." looks outdated now.

moulin/build_generator.py
svlad-90: Hi @lorc, done.

moulin/build_generator.py
svlad-90: Hi @lorc, done.

moulin/builders/custom_script.py
svlad-90: Hi @lorc, done.

moulin/builders/yocto.py
svlad-90: Hi @lorc, done.

moulin/builders/zephyr.py
svlad-90: Hi @lorc, done.

moulin/build_generator.py
svlad-90: Hi @lorc, done.

moulin/build_generator.py
svlad-90: Hi @lorc, done.

moulin/builders/zephyr.py
svlad-90: Hi @lorc, done.

moulin/builders/zephyr.py
svlad-90: Hi @lorc, done.

moulin/builders/zephyr.py
svlad-90: Hi @lorc, done.

moulin/utils.py
svlad-90: Hi, @lorc.

This is intentional backward compatibility. construct_fetcher_dep_cmd() was the producer of new --fetcherdep calls, so it was replaced to ensure newly generated build files use --dep.

The --fetcherdep handler is still needed as a legacy consumer: existing build.ninja files generated by an older Moulin version embed a --fetcherdep call. Updating Moulin does not automatically regenerate those files because the Moulin executable is not a dependency of the regeneration rule. Such files can therefore invoke the new executable with the old option.

=> In other words, new files emit only --dep, while the CLI temporarily accepts both formats. It would be better to remove --fetcherdep entirely several releases later (not now) to avoid build failures in existing local builds.

lorc: Okay, I see your point. But IMO, it does not justify all the complexity in the code. Just make the following: if moulin sees a `--fetcherdep` call, it prints something like "build.ninja was created with older version of moulin. Please re-run moulin, to update the build file"

moulin/build_generator.py
```diff
@@ -19,6 +20,16 @@
 BUILD_FILENAME = 'build.ninja'
 
 
+@dataclass
+class DependencyContext:
+    build_dir: str
+    component_node: Any
```
lorc: Pleasy, no `Any`. This is very bad idea. I want typing to be as tight as possible

lorc (changes requested):
Could you please run `mypy` locally to ensure that typing is solid?

moulin/utils.py
svlad-90: Hi @lorc, done.

moulin/build_generator.py
svlad-90: Hi @lorc, done.

lorc (approved):
`Reviewed-by: Volodymyr Babchuk <volodymyr_babcuk@epam.com>`

PR #152, opened 2026-07-15
yocto: pass base distro to layer sync utility & show captured subprocess errors

svlad-90:
yocto: pass base distro to layer sync utility
Yocto build rules already source oe-init-build-env through the configured
base_distro value. The layer sync utility did not receive that value and
always sourced poky/oe-init-build-env internally.

This breaks OpenEmbedded-core based layouts, where the init script lives
under openembedded-core/oe-init-build-env and there is no poky directory.
Pass the generated distro_dir variable into the utility and keep poky as
the command-line default for compatibility with direct utility calls.

----

yocto: show captured subprocess errors
Yocto layer sync runs bitbake-layers show-layers with capture=True
when a managed-layers stamp already exists. If BitBake rejects the
current bblayers.conf, for example because it references a missing
layer, subprocess raises CalledProcessError but the captured
stdout/stderr are not printed.

As a result Moulin reports only the Python traceback and hides the
actionable Yocto diagnostic. Print captured output before re-raising so
the command still fails with the same behavior, but the real BitBake
error is visible in the build log.

### DIFF
--- a/moulin/builders/yocto.py
+++ b/moulin/builders/yocto.py
@@ -340,12 +340,20 @@ def _env_prefix(yocto_dir: str, work_dir: str) -> str:
 
 def _run_bash(cmd: str, *, capture=False) -> subprocess.CompletedProcess:
     """Run a bash -lc command."""
-    return subprocess.run(
-        ["bash", "-lc", cmd],
-        check=True,
-        capture_output=capture,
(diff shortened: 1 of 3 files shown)


lorc (approved):
Thanks!

Issue #125, opened 2025-02-13
Ninja doesn't see changes in the product's layers


rshym:
As far as the product's layers are not specified in the yaml section `sources`, they are not provided as dependency for the ninja.
As a result, the ninja doesn't see changes in the `product/layers/*` and doesn't start the build of the modified component.

This issue can be checked by `ninja -d explain` and `ninja -t deps`.

rshym:
The proposed solution is to add a new section into the yaml: `external-sources:` or `prefetched-sources:`. Directories/files listed under this section will not be fetched (`fetch-...` target will not be created for them), but they will be added to the list of ninja's dependencies.
Aside from solving the original issue, this may be used for adding external binaries like prebuilt GSX, etc.

lorc:
Fixed with #149

PR #154, opened 2026-07-16
yocto: exclude work dir from layer dependency scan

svlad-90:
Yocto layer dependency tracking walks configured layer directories after a build and writes the result to .moulin_<component>.d.

When a broad layer path also contains the active Yocto work directory, generated files such as buildhistory can be recorded as Ninja inputs and make the depfile huge.

Skip the active work directory while walking layer dependencies, but keep the existing work_dir='.' behavior where the work directory is the Yocto root itself.

### DIFF
--- a/moulin/builders/yocto.py
+++ b/moulin/builders/yocto.py
@@ -186,6 +186,12 @@ your YAML Moulin configuration files.
     return result
 
 
+def _is_same_or_child_path(child_candidate: str, parent_candidate: str) -> bool:
+    child_candidate = os.path.abspath(os.path.normpath(child_candidate))
+    parent_candidate = os.path.abspath(os.path.normpath(parent_candidate))
+    return os.path.commonpath([child_candidate, parent_candidate]) == parent_candidate
+
+
 class YoctoBuilder:
     """
     YoctoBuilder class generates Ninja rules for given build configuration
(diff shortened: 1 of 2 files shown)


moulin/builders/yocto.py
```diff
@@ -304,13 +309,44 @@ def get_extra_dep_file_list(self) -> List[str]:
 
         deps: List[str] = []
         layers_base = os.path.join(self.yocto_dir, self.work_dir)
+        yocto_dir = os.path.abspath(os.path.normpath(self.yocto_dir))
+        work_dir = os.path.abspath(os.path.normpath(os.path.join(self.yocto_dir, self.work_dir)))
```
lorc, line 313: You are making assumption that `self.work_dir` is a relative path. Do you have any guarantees that it is always a case?

moulin/builders/yocto.py
```diff
@@ -304,13 +309,44 @@ def get_extra_dep_file_list(self) -> List[str]:
...
         layers = _filter_yocto_core_layers(_flatten_layers(layers_node))
         for layer in layers:
             layer_path = os.path.normpath(os.path.join(layers_base, layer))
             if not os.path.isdir(layer_path):
                 raise YAMLProcessingError(f"Can't find Yocto layer directory '{layer_path}'",
                                           layers_node.mark)
-            for dirpath, _, filenames in os.walk(layer_path):
+
+            # Edge case: if a layer points inside the build dir, os.walk() starts
+            # below the pruning point, so skip it before walking.
```
lorc: Will this break the very use case, why we have this functionality in the first place?

moulin/builders/yocto.py
```diff
@@ -304,13 +309,44 @@ def get_extra_dep_file_list(self) -> List[str]:
...
+            layer_is_inside_work_dir = (work_dir_can_be_pruned
+                                        and _is_same_or_child_path(layer_path, work_dir))
+            if layer_is_inside_work_dir:
```
lorc: Please make sure that this change does not break https://github.com/xen-troops/moulin/issues/125

(you can check https://github.com/xen-troops/meta-xt-prod-devel-rpi5 as a real world example)

moulin/builders/yocto.py
svlad-90: Hi @lorc, this should cause no issue.

---

https://docs.python.org/3/library/os.path.html

Join one or more path segments intelligently. The return value is the concatenation of path and all members of *paths, with exactly one directory separator following each non-empty part, except the last. That is, the result will only end in a separator if the last part is either empty or ends in a separator.

If a segment is an absolute path (which on Windows requires both a drive and a root), then all previous segments are ignored and joining continues from the absolute path segment. On Linux, for example:

```
os.path.join('/home/foo', 'bar')
'/home/foo/bar'
os.path.join('/home/foo', '/home/bar')
'/home/bar'
```

---

If work_dir is a relative path, it will be preceded by yocto_dir. If it is absolute, it will be used directly.

moulin/builders/yocto.py
svlad-90: Checked. This should not break #125.

The implementation does not blanket-skip layers under `work_dir`. It prunes `work_dir` only when the Yocto build directory is nested beneath a broader scanned layer path. An explicitly configured layer inside `work_dir` is still scanned and tracked; this is covered by `test_yocto_layer_deps_keep_layer_nested_under_work_dir`.

  I also checked the RPi5 YAML as a real-world example. Its Yocto layer entries resolve to specific layer directories, for example:

```
yocto/meta-xt-common/meta-xt-domu
layers/xt-prod-devel-rpi5-domu
etc.
```

They do not resolve to the Yocto build directory itself, nor to a parent directory that contains it.

moulin/builders/yocto.py
svlad-90: Good point. #125 is about product-local layers, not about generated files under the Yocto build directory.

But I agree that this layout assumption is too broad. I've changed the code to keep the scanning of the layer directory inside the Yocto build directory.

After the update, only the configured Yocto `work_dir` subtree is pruned when it appears under a broader scanned path. Explicit layers inside `work_dir` are still tracked as well, covered by `test_yocto_layer_deps_keep_layer_nested_under_work_dir`.

lorc (approved):

PR #156, opened 2026-07-17
west: handle group filters starting with dash

svlad-90:
Pass a '--' separator before the generated manifest.group-filter value so exclusion-style filters are not parsed as west config options.

Extend the west fetcher group_filter integration test to cover a value that starts with '-'.


lorc:

PR #157, opened 2026-07-18
yocto: prune Yocto generated and .git dirs from deps

svlad-90:
Product layers can contain generated build trees. They can also be reached through symlinked layer paths. This can create huge depfiles and make Ninja fail.

Query BitBake generated directory variables with bitbake-getvar. Prune matching trees before os.walk descends into them, and skip layer-local .git directories so repository metadata is not recorded in depfiles.

Missing variables stay warning-only for BitBake compatibility.

### DIFF
--- a/moulin/builders/yocto.py
+++ b/moulin/builders/yocto.py
@@ -20,6 +20,22 @@ from pathlib import Path
 
 
 YOCTO_CORE_LAYERS = ["../poky/meta", "../poky/meta-poky", "../poky/meta-yocto-bsp", "../openembedded-core/meta"]
+# Moulin keeps the variable list, but BitBake provides the actual paths. TOPDIR
+# is set by BitBake; the other variables are OE-Core/Poky build directories
+# normally defined from TOPDIR/TMPDIR in meta/conf/bitbake.conf. BUILDHISTORY_DIR
+# is optional and exists only when buildhistory metadata is enabled.
+YOCTO_GENERATED_DIR_VARIABLES = [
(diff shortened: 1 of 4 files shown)


andr2000:

moulin/builders/yocto.py
```diff
@@ -188,11 +204,77 @@ def _filter_yocto_core_layers(layers: List[str]) -> List[str]:
 
 
 def _is_same_or_child_path(child_candidate: str, parent_candidate: str) -> bool:
-    child_candidate = os.path.abspath(os.path.normpath(child_candidate))
-    parent_candidate = os.path.abspath(os.path.normpath(parent_candidate))
+    # realpath is required for product layers reached through symlinks, e.g.
```
lorc, line 207: I believe, this is a fix for a separate issue, which should go into separate commit

lorc (changes requested):
Honestly, I don't understand how we got here and why you need more and more code to fix the original problem.

Like, the original problem was "Ninja does not track changes in layer, if it didn't pulled it from git". How we come to situation where you are traversing the whole work directory and then adding more and more heuristics to skip unrelated files?

All you need is to query bitbake for list of layers that are used for build and scan these layers. That's all.

svlad-90:
After discussion with @lorc it was agreed that coverage of such use-cases like "I have a Yocto working directory inside the meta-layer" is not an option. The product's layout limitations will be listed in the documentation in a separate PR.

I'm closing this one.

PR #159, opened 2026-07-18
custom_script: add inline script support

svlad-90:
Allow small custom script bodies to be embedded directly in a Moulin configuration with inline_script. Keep script as the external script path mode, and reject configurations that set both or neither option.

Write inline_script content to a generated script file under work_dir and invoke it through bash before running post-build dependency generation. This lets YAML use normal shell quoting without feeding the script body through the Ninja rule command line.

Document the new option and add integration coverage for quotes, a trap, a subshell, target generation, and config-file argument passing.


andr2000:

moulin/builders/custom_script.py
```diff
@@ -31,6 +33,9 @@ def gen_build_rules(generator: ninja_syntax.Writer):
         "( [ -d $work_dir ] || mkdir $work_dir )",
         "echo '# Code generated by moulin. All manual changes will be lost' > $config_file",
         "echo $b64config | base64 -d - >> $config_file",
+        "if [ -n \"$inline_b64_script\" ]; then "
```
lorc: Do not extend existing build rule. It is already looks complex enough. Add a new rule, tailored for inline_script specifically

moulin/builders/custom_script.py
```diff
@@ -31,6 +33,9 @@ def gen_build_rules(generator: ninja_syntax.Writer):
...
+        "echo $inline_b64_script | base64 -d - > "
+        "$quoted_inline_script_file_name; fi",
```
lorc: What is `$quoted_inline_script_file_name` is already exist in the directory.  Use `tmpfile` instead

moulin/builders/custom_script.py
```diff
@@ -68,11 +73,31 @@ def gen_build(self):
...
+            raise YAMLProcessingError("Either 'script' or 'inline_script' is required",
+                                      self.conf.mark)
+
+        inline_script_file_name = os.path.join(work_dir, f"{self.name}.sh")
+        if inline_script_node:
+            script_cmd = f"bash {shlex.quote(inline_script_file_name)}"
+            # Store the multiline inline script in a Ninja variable without
+            # exposing its shell syntax to the generated command.
+            script_bytes = inline_script_node.as_str.encode("utf-8")
+            inline_b64_script = base64.b64encode(script_bytes).decode("utf-8")
```
lorc: Why do you need this black magic with storing b64-encoded data in Ninja files, when Misha added API to cover cases like this?

https://github.com/xen-troops/moulin/pull/132/changes/472ac4674aa6ef4a9cac668be8e81847f25edc5e

lorc (changes requested):

moulin/builders/custom_script.py
svlad-90: Hi @lorc, done. An additional 'cs_inline_build' rule was added.

moulin/builders/custom_script.py
svlad-90: Hi @lorc, done. I've switched the implementation to use a tempfile.

moulin/builders/custom_script.py
svlad-90: Hi @lorc, done. I've switched the implementation to the use of the utility dispatch mechanism.

moulin/builders/custom_script.py
```diff
@@ -50,6 +56,83 @@ def gen_build_rules(generator: ninja_syntax.Writer):
...
+                encoding="utf-8",
+                prefix=".moulin-inline-",
+                suffix=".sh",
+                dir=work_dir,
+                delete=False) as stream:
+            script_file = stream.name
+            stream.write(inline_script)
+
+        os.chmod(script_file, 0o700)
+        result = subprocess.run(["bash", script_file, *_get_inline_script_args(conf), config_file])
```
lorc: Whoa, not so fast. Non-inline variant support running any scripts (or any programs, honestly). Even documentation shows how to run a Python script.  And here you are limiting functionality to bash scripts only with no apparent reason.

lorc (changes requested):
Please ensure that user can use any scripting language they want

moulin/builders/custom_script.py
svlad-90: Hi @lorc, done.

lorc (approved):

PR #160, opened 2026-07-19
Revert "yocto: exclude generated work dir from layer deps" + documentation update

svlad-90:
Revert "yocto: exclude generated work dir from layer deps"
This reverts commit https://github.com/xen-troops/moulin/commit/77356894315bab7c22930f866c5ad19d3021f9dc.

The layout handled by that change is not a supported product layout.
Moulin expects the Yocto build directory to live outside managed layer
directories. If a configured layer contains work_dir, generated BitBake
output can be tracked as layer source dependencies. Products should avoid
that layout instead of relying on Moulin to prune generated files while
walking a layer directory.

Drop the work_dir pruning logic and the tests that documented the
unsupported nested-build-dir behavior.

----

yocto: reject unsupported layer work_dir layouts
Yocto layer dependency tracking records files under the directories
listed in the builder's layers option. A product layout where one of
those listed layer directories also contains the Yocto work_dir makes
generated BitBake output, such as tmp, cache, or buildhistory files,
look like layer source dependencies.

Raise YAMLProcessingError when the configured work_dir resolves inside
a managed layer path. Resolve symlinks before comparing paths, so
layouts that reach the same physical directory through a layer symlink
are rejected before build.ninja is written.

Keep layer dependency scanning from traversing symlinked directories
explicitly. This preserves source dependency tracking for files
physically under the layer while avoiding generated work_dir contents
reached through a layer symlink.

Document both the guard and its limits. Moulin can reject the
configured nested work_dir layout, but it does not try to detect every
way generated output can be copied, redirected, or otherwise written
into a layer directory.


lorc (approved):

andr2000:
If this is not supported, so can we raise an exception if we hit this limitation?

svlad-90:
Hi @andr2000,

I agree with your point. Waiting for 2 hours to understand that your layout is unsupported during deps file processing is also unacceptable.

I've added an early guard for the exact unsupported layout where the configured Yocto work_dir is located under one of the configured layer paths. That is straightforward and does not require restoring the pruning logic.

**However, this would only catch the structural case Moulin can infer from its configuration. Nothing more than that.** It would not reliably detect every way generated files can end up under a layer: BitBake variables may redirect output directories elsewhere, other build systems ( Zephyr's west call :) ) may write to the layer tree if configured to do so, users may copy generated files manually, and symlinks can cause physical paths to differ from configured paths.

@lorc, updated implementation checked on @andr2000's scenario. It still follows your directive not to make endless workarounds. Please take a look once again.

lorc (approved):

PR #161, opened 2026-07-21
Yocto populate sdk

svlad-90:
yocto: add populate_sdk builder option

Allow Yocto components to request BitBake SDK generation by setting
populate_sdk to true. The default build path still invokes the
configured build_target directly.

When populate_sdk is enabled, pass -c populate_sdk together with the
build target through the generated Ninja rule. Keep target_images as the
declared outputs so callers can point Moulin at the expected SDK
installer path.

Document the new option and cover the generated bitbake arguments in the
Yocto builder unit tests.

----

version: bump version to 0.32

Taking into account added Yocto populate_sdk builder option and other
changes, bump the version up.

### DIFF
--- a/moulin/builders/yocto.py
+++ b/moulin/builders/yocto.py
@@ -97,6 +97,20 @@ def gen_build_rules(generator: ninja_syntax.Writer):
                    deps="gcc",
                    depfile=".moulin_$name.d",
                    restat=True)
+    generator.newline()
+
+    cmd = " && ".join([
+        "pushd $yocto_dir > /dev/null",
+        "source $distro_dir/oe-init-build-env $work_dir",
+        "bitbake -c populate_sdk $target",
+        "popd > /dev/null",
+        "touch $out",
+    ])
+    generator.rule("yocto_populate_sdk",
+                   command=f'bash -c "{cmd}"',
+                   description="Yocto Populate SDK: $name",
+                   pool="console",
+                   restat=True)
 
 
 def _flatten_yocto_conf(conf: YamlValue) -> List[Tuple[str, str]]:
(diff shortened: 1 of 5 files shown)


moulin/builders/yocto.py
```diff
@@ -317,12 +292,16 @@ def gen_build(self):
         self.generator.newline()
 
         # Next step - invoke bitbake. At last :)
+        bitbake_args = self.conf["build_target"].as_str
```
lorc:
```
        bitbake_args = [self.conf["build_target"].as_str]
        if self.conf.get("populate_sdk", False).as_bool:
            bitbake_args.append("-c populate_sdk")
```

moulin/builders/yocto.py
```diff
@@ -317,12 +292,16 @@ def gen_build(self):
...
+        if self.conf.get("populate_sdk", False).as_bool:
+            bitbake_args = "-c populate_sdk " + bitbake_args
+
         targets = self.get_targets()
         self.generator.build(targets,
                              "yocto_build",
                              local_conf_target,
                              variables=dict(common_variables,
-                                            target=self.conf["build_target"].as_str,
+                                            bitbake_args=bitbake_args,
```
lorc: `bitbake_args=shlex.join(bitbake_args)`

lorc (changes requested):
There are more pythonic ways to operate with lists ;)


With that fixed:

andr2000:
`boolean flag that runs BitBake's SDK population task instead of the default build task`
can we run not instead, but as an additional task?

moulin/builders/yocto.py
svlad-90: Hi @lorc, done. But I've used '''bitbake_args.extend(["-c", "populate_sdk"])''' to avoid passing "-c populate_sdk" as a single argument after ```shlex.join(bitbake_args)```.

moulin/builders/yocto.py
svlad-90: Hi @lorc, done.

lorc:
IIRC, the resulting behavior will be the same

svlad-90:
Hi @lorc, @andr2000, done.

No-SDK ninja graph:
domd-no-sdk-ninja-graph

Populate SDK ninja graph:
domd-with-sdk-ninja-graph

andr2000:
LGTM WRT logic implemented

moulin/builders/yocto.py
```diff
@@ -86,7 +86,7 @@ def gen_build_rules(generator: ninja_syntax.Writer):
     cmd = " && ".join([
         "pushd $yocto_dir > /dev/null",
         "source $distro_dir/oe-init-build-env $work_dir",
-        "bitbake $target",
+        "bitbake $bitbake_args",
```
lorc: You don't need this change anymore

moulin/builders/yocto.py
```diff
@@ -317,15 +306,27 @@ def gen_build(self):
         self.generator.newline()
 
         # Next step - invoke bitbake. At last :)
+        build_target = self.conf["build_target"].as_str
```
lorc: .. and this as well

lorc (changes requested):
Please remove unnecessary changes

moulin/builders/yocto.py
svlad-90: Hi @lorc, done.

